PR #24, opened 2012-12-14
Support for specificity rules and media queries

theospears:
A few changes to try to apply rules in the correct order based on specificity. It also prevents inlining style tag which have media queries on them. There's also a tweak to allow css selector caching to speed things up when sending many emails.


peterbe:
So, can we close this now?

theospears:
The changes from this I really cared about went in with #30 and as you say this is very out of date, so I'll go ahead and close.

PR #105, opened 2015-02-14
Caching

michaeljones:
Hi,

I have a production use of premailer that is going quite slowly. Looking at the code, I thought to introduce some caching to avoid re-parsing the CSS text. I've introduced a test case which is a little bulkier than your other ones and added a test which checks the result and the time required to transforms the html.

Without caching it takes 16.4 seconds.
With caching it takes 1.33 seconds.

I think this is a good addition to the code and, though I'm aware that caching is a space vs time trade off, I don't think that the extra memory in this situation should be a serious concern. I haven't done any tests to prove that though.

The commit `Add compare method to test` attempts to resolve an unused variable warning from one of the tests. It is unrelated to the main changes and can easily be dropped if I misread the situation.

Cheers,
Michael

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -48,9 +48,7 @@ class ExternalNotFoundError(ValueError):
 grouping_regex = re.compile('([:\-\w]*){([^}]+)}')
 
 
-styles_cache = {}
-
-def merge_styles(old, new, class_=''):
+def merge_styles(old, new, styles_cache, class_=''):
     """
     if ::
       old = 'font-size:1px; color: red'
@@ -389,6 +387,7 @@ class Premailer(object):
 
         first_time = []
         first_time_styles = []
+        styles_cache = {}
         for __, selector, style in rules:
             new_selector = selector
             class_ = ''
@@ -405,11 +404,11 @@ class Premailer(object):
             for item in sel(page):
                 old_style = item.attrib.get('style', '')
                 if not item in first_time:
-                    new_style = merge_styles(old_style, style, class_)
+                    new_style = merge_styles(old_style, style, styles_cache, class_)
                     first_time.append(item)
                     first_time_styles.append((item, old_style))
                 else:
-                    new_style = merge_styles(old_style, style, class_)
+                    new_style = merge_styles(old_style, style, styles_cache, class_)
                 item.attrib['style'] = new_style
                 self._style_to_basic_html_attributes(item, new_style,
                                                      force=True)
@@ -419,7 +418,7 @@ class Premailer(object):
             old_style = item.attrib.get('style', '')
             if not inline_style:
                 continue
-            new_style = merge_styles(old_style, inline_style, class_)
+            new_style = merge_styles(old_style, inline_style, styles_cache, class_)
             item.attrib['style'] = new_style
             self._style_to_basic_html_attributes(item, new_style, force=True)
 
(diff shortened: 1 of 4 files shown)


michaeljones:
I'll attempt to address the failing tests tomorrow. It seems mostly be to be because the Travis test hosts are understandably not packing the performance of my own machine. I've no idea how to deal with PyPy and how fast or slow that should be. 

Maybe I should drop the timing aspect of the tests?

ktoh:
+1 @michaeljones excellent contribution.  I noticed the same problems with slowness and was just about to look into a caching solution.

michaeljones:
Tests are fixed with a little back and forth. Hope it looks acceptable.

Thanks for the positive feedback @ktoh!

peterbe:
This is great stuff. Optimization is important. 
However, a couple of points:
a) The new test does not belong in the unit tests. It's a benchmark. It can't be part of the test framework. Feel free to create a sub-directory called `benchmarking` or something and in it put that sample HTML, a README and a python script that imports premailer and runs it. 
b) I started a branch that uses a simple memoize decorator on those crucial functions. The results I get are:

| Branch       | Time (100 iterations) |
|--------------|-----------------------|
| master       | 99.4417426586         |
| memoize      | 36.1815743446         |
| michealjones | 55.2455043793         |

So, that's ~2.5 times faster just by inserting two little decorators. However, it's still ~0.36 seconds per page. I used a 18Kb HTML page. A sample based in Ink by Zurb. 

My branch currently doesn't pass tests. It's because the memory is stick between tests. So, if I can only find a way to undo that each time, things could get interesting.

michaeljones:
Hi,

Thank you for the response. I am more than happy to act on your comments and move things around if you feel that it is worth while? I'm a little confused as if your branch is faster, is it worth continuing mine?

Cheers,
Michael

vanng822:
@peterbe please take look on my pull request #107 It does not make much different when you run on small number of selector and elements, but it make different when you have bigger number

peterbe:
I did some more work and all tests pass and instead of 36s for 100 iterations, it's now down to 31s.

peterbe:
@michaeljones I think @vanng822 's PR blows away the problem. It really does work and passes all the tests. If you don't mind, I'm going to close this PR. Let's all head over to #107 and get that landed.

michaeljones:
Cool, thanks for the update. Glad it is getting faster.

peterbe:
So am I. I bet there might be a bunch of other places where we can do "cheap" optimizations too.

vanng822:
There is more to optimise I think such as keep css declaration so you don't have to run thread lock (csstext_to_pairs) too often. I think it is expensive when you have a server where it's task is sending email. I was thinking of taking a look on it next.

michaeljones:
Thank you very much for the releases and sorry to keep abusing this pull request as a chat forum, but I have one more question.

Would you mind making git tags for the PyPI releases? They are reasonably clear from the 'bump to x.x.x' style commits but it seems reasonable (to me!) to tag them as well.

peterbe:
@michaeljones thanks for the thought but at the moment, no. 
With using regular git commits I can make travis build on every push and try to update pypi every time which only works if the version has changed. 
This way, there's only one place to worry about updating the version number and that's in `premailer/__init__.py` and nothing else.

PR #107, opened 2015-02-17
Performance improvement of inline styling

vanng822:
We were entering problem with our emails which required 15-60s to process most of them and cpu usage was high, 100% cpu during those seconds. They take now around 0.4-1.5 seconds to process

After investigation I realize that inline styling didn't have a linear curve on time vs elements/classes. It multiply, I can not say what power is it but definitive some kind of exponential. For small amount it fast but when the mail is big and have complex rules it takes to much time.

Instead of converting rules to string and inserting into style-attribute and take it back again, parsing it with cssutils, merge and insert back to style-attribute, and repeat for many many run... (This is where time is multiplying) I collects all elements that we need to apply the rules on and apply rules for each of them in one goal.

Further more, our css are long and parse and go through rules on each of them take around 1-1.5 seconds. This is the reason I added a cache to gain around 80% of this time. It can be turned off for those who don't want it.

Here is a example code (thanks michaeljones) where the old code took 185.72238493s and new code took 5.29419207573s

https://gist.github.com/vanng822/ff706586bf6a91f3432d

Please take a look
Thank you

### DIFF
--- a/premailer/cache.py
+++ b/premailer/cache.py
@@ -19,7 +19,22 @@ class _Cache(object):
         self.cache = {}
 
 def function_cache(expected_max_entries=1000):
-    
+    """
+        function_cache is a decorator for caching function call
+        the argument to the wrapped function must be hashable else
+        it will not work
+        
+        expected_max_entries is for protecting cache failure. If cache misses more than
+        this number the cache will turn off itself. Specify None you sure that the cache
+        will not cause memory limit problem
+        
+        Args:
+            expected_max_entries(integer OR None): will raise if not correct
+            
+        Returns:
+            function
+            
+    """
     if expected_max_entries is not None and not isinstance(expected_max_entries, int):
         raise TypeError('Expected expected_max_entries to be an integer or None')
     
--- a/premailer/cache.py
+++ b/premailer/cache.py
@@ -1,0 +1,58 @@
+import functools
+
+class _HashedSeq(list):
+    # # From CPython
+    __slots__ = 'hashvalue'
+
+    def __init__(self, tup, hash=hash):
+        self[:] = tup
+        self.hashvalue = hash(tup)
+
+    def __hash__(self):
+        return self.hashvalue
+
+## if we only have nonlocal
+class _Cache(object):
+    def __init__(self):
+        self.off = False
+        self.missed = 0
+        self.cache = {}
+
+def function_cache(expected_max_entries=1000):
+    
+    if expected_max_entries is not None and not isinstance(expected_max_entries, int):
+        raise TypeError('Expected expected_max_entries to be an integer or None')
+    
+    # indicator of cache missed
+    sentinel = object() 
+    
+    def decorator(func):
+        cached = _Cache()
+    
+        @functools.wraps(func)
+        def inner(*args, **kwargs):
+            if cached.off:
+                return func(*args, **kwargs)
+            
+            keys = args
+            if kwargs:
+                sorted_items = sorted(kwargs.items())   
+                for item in sorted_items:
+                    keys += item
+                    
+            hashed = hash(_HashedSeq(keys))
+            result = cached.cache.get(hashed, sentinel)
+            if result is sentinel:
+                cached.missed += 1
+                result = func(*args, **kwargs)
+                cached.cache[hashed] = result
+                # # something is wrong if we are here more than expected
+                # # empty and turn it off
+                if expected_max_entries is not None and cached.missed > expected_max_entries:
+                    cached.off = True
+                    cached.cache.clear()
+                    
+            return result
+            
+        return inner
+    return decorator
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -32,6 +32,7 @@ import cssutils
 from lxml import etree
 from lxml.cssselect import CSSSelector
 from premailer.merge_style import merge_styles, csstext_to_pairs
+from premailer.cache import function_cache
 
 __all__ = ['PremailerError', 'Premailer', 'transform']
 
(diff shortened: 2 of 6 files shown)


peterbe:
Awesome! I can verify that this is indeed a big improvement. 

| Branch       | Time (100 iterations) |
|--------------|-----------------------|
| master       | 99.4417426586         |
| memoize      | 36.1815743446         |
| vanng822 | 16.0634069443        |

That means that your code is over **6 times faster**. 

I'm going to start picking through it for a merge.

premailer/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import, unicode_literals
 from .premailer import Premailer, transform
 
-__version__ = '2.8.1'
+__version__ = '2.8.2'
```
peterbe: Would you mind reverting this one? I don't like to make a new release by merging a PR. It should be done as a separate process.

premailer/cache.py
```diff
@@ -0,0 +1,73 @@
...
+        return self.hashvalue
+
+## if we only have nonlocal
+class _Cache(object):
+    def __init__(self):
+        self.off = False
+        self.missed = 0
+        self.cache = {}
+
+def function_cache(expected_max_entries=1000):
```
peterbe: Was this function written specifically for premailer or was it copy-n-pasted from somewhere else?

premailer/premailer.py
```diff
@@ -161,7 +106,8 @@ def __init__(self, html, base_url=None,
                  method="html",
                  base_path=None,
                  disable_basic_attributes=None,
-                 disable_validation=False):
+                 disable_validation=False,
+                 cache_css_parsing=True):
```
peterbe: Thank you! I think this is a crucial option. It's interesting that you chose to make it on by default.

peterbe:
@vanng822 Also, would you mind rebasing and squashing some of the minor commits (for example "Fix invalid syntax for python26") into one big commit? 

I see a bunch of pep8 violations. Strangely some of the changes you made were pep8 fixes. Do you mind fixing some of them up first? If not, I can do that later.

premailer/__init__.py
vanng822: Will do, It was for our rpm to build. Sorry about it

premailer/cache.py
vanng822: It was written for premailer if you are worry about copyright. The hash class is taken from Cpython thought.

I tried out a couple of options but not happy with them. Found functools.lru_cache but was only for python3

premailer/premailer.py
vanng822: Yes, it was a hard decision to put True on that kwarg. I always took other to not break the api when adding new stuffs. But this time I added true since this is mostly silent and the chance that someone could have problem with memory is small. But they will have some improvement.

We can set to False if you like but I think it may worth the risk

vanng822:
Sorry about a lot of fix and tricks. Tried to run tox on my computer but it could install pacakges :-(

Let see if I can figure out how to fix else I will need your expertis

vanng822:
It seems not easier to rebase my master and keep the PR, unless I do push force after rebase. Do you have any tips?

What I could do is rebase and push to a new branch and create a new PR? :-/

peterbe:
You're actually expected to do a force push. It's ok to do force pushes as long as they're not to origin/master. But it's OK to do it to your own branch.

premailer/cache.py
peterbe: Excellent. Yes I was worried about copyright. The whole project is Python which protects us but since premailer is used inside corporate projects it's better to be safe than sorry.

premailer/premailer.py
peterbe: You did a nice test that checked that it's possible to switch off the cache. Let's leave it on.

peterbe:
@vanng822 Would you be interested in sharing your company's name and logo into the README as a "Premailer is used by..." section?

vanng822:
Ok, then I will do a rebase and force push.

Regarding logo, any preferred spec? allowed to link to the site?

peterbe:
@vanng822 Can you email on on mail@peterbe.com and I'll make the necessary change to the README and perhaps get some details from you about your company.

peterbe:
I'm restarting the travis build.

vanng822:
Thanks!

Now I rebased and run the tests on python 2.7 for each of the commits. All passed. It may be so that if someone check out one revision and try to run on it, it may not be really good. But I don't want to remove all hístory and no one should do that :-p

peterbe:
Hurray!

vanng822:
Yay!
Thank you!

peterbe:
I might do a general pep8 cleanup and introduce some pep8 and pyflake checking as part of the testing.

vanng822:
Oh, that one, yes please. I tried on my eclipse but it didn't help me much :-(
Would be nice to have a git hooks somehow. I didn't notice that python 2.6 is supported until the test failed on build

michaeljones:
Hi,

Are you planning a new release with these performance improvements? Is there any more work to be done?

Cheers,
Michael

vanng822:
Hi, I tried to optimize on _parse_style_rules where I replaced join_css_properties with building a list off tuple like csstext_to_pairs. It is only 10% improvement. I did it quickly, by just applying on bulk_important and bulk_normal,  to see how much we can gain but this number is quite small. The time spent to make everything consistent and not introducing new bugs is kind of too much

So for me it is done for this round.

peterbe:
@michaeljones the 2.8.2 release; coming to a pypi near you very soon.

michaeljones:
Excellent! Thanks for keeping us updated.

PR #116, opened 2015-03-22
Disable pretty-printing by default for CLI

luhn:
I don't see much reason for pretty-printing, it just introduces new whitespace that I've found to be problematic.  This patch disables pretty-pretty in the CLI, but it can be reenabled with the `--pretty` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ You can also use premailer from the command line by using his main module.
     --disable-basic-attributes DISABLE_BASIC_ATTRIBUTES
                           Disable provided basic attributes (comma separated)
     --disable-validation  Disable CSSParser validation of attributes and values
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -120,6 +120,7 @@ You can also use premailer from the command line by using his main module.
     --disable-basic-attributes DISABLE_BASIC_ATTRIBUTES
                           Disable provided basic attributes (comma separated)
     --disable-validation  Disable CSSParser validation of attributes and values
+		--pretty              Pretty-print the outputted HTML.
```
peterbe: There's some strange indentation going on there. Both the `--pretty` and the `Pretty-print the...` are over-indented.

luhn: Whoops.  Didn't notice that.  Fixed.

Issue #100, opened 2014-12-26
Can't handle &nbsp;


loentar:
If source html contain `&nbsp;` exception will thrown:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/local/lib/python2.7/dist-packages/premailer/__main__.py", line 133, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/local/lib/python2.7/dist-packages/premailer/__main__.py", line 128, in main
    options.outfile.write(p.transform())
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 45: ordinal not in range(128)
```

Test html:

```
<html>
  <body>&nbsp;</body>
</html>
```

peterbe:
I started writing a unit test to reproduce this. So I wrote this:

```python
    def test_nbsp_only(self):
        html = """
        <html>
          <body>&nbsp;</body>
        </html>
        """
        p = Premailer(html)
        p.transform()
```

It didn't fail. 

Any other ideas of how to reproduce this?

kengruven:
The Python Premailer class supports "&nbsp;", but as part of its transformation, it converts it to the character U+00A0 (or maybe the UTF-8 bytes C2 A0).

If you run it in Python to convert a string to a string (in memory), it works.  If you run it via the command-line using premailer/__main__.py, it fails:

    python -m premailer --file /tmp/nbsp.html --output /tmp/nbsp-pm.html
    (traceback...then:)
    UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 29: ordinal not in range(128)

It looks like the output file is being created with an ASCII-only codec.

peterbe:
What does that `/tmp/nbsp.html` look like? E.g. if you do:
```python
>>> import codecs
>>> codecs.open('/tmp/nbsp.html', 'r', 'utf-8').read()
```

kengruven:
It's plain ASCII.  It's the "Test html" snippet from loentar's original issue, at top.

Running your command with Python 2.7 gives me:

    u'<html>\n  <body>&nbsp;</body>\n</html>\n'

peterbe:
[...]

PR #120, opened 2015-04-17
Always use UTF-8 for writing files, when running from the command line.

kengruven:
Python 2.x argparse.FileType doesn’t have an encoding= keyword parameter, so just re-open the file.

It’s not pretty, but it works.  If supporting only Python 3.4 and up, we can remove all this, and use argparse.FileType(‘w’, encoding=‘UTF-8’) above instead.

See bug #100.

### DIFF
--- a/premailer/__main__.py
+++ b/premailer/__main__.py
@@ -117,8 +117,9 @@ def main(args):
 
     if options.outfile is not sys.stdout:
         options.outfile.close()
-        options.outfile = codecs.open(options.outfile.name, 'w',
-                                      encoding='UTF-8')
+        options.outfile = codecs.open(
+            options.outfile.name, 'w', encoding='UTF-8'
+        )
 
     html = options.infile.read()
     if hasattr(html, 'decode'):  # Forgive me: Python 2 compatability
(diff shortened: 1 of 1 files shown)


premailer/__main__.py
```diff
@@ -114,6 +115,11 @@ def main(args):
             options.disable_basic_attributes.split()
         )
 
+    if options.outfile is not sys.stdout:
+        options.outfile.close()
+        options.outfile = codecs.open(options.outfile.name, 'w',
```
peterbe: Can you indent this like this:
```python
options.outfile = codecs.open(
    options.outfile.name, 'w', encoding='UTF-8'
)
```
That way the second lines massive line of spaces isn't dependent on the length of the first line.

peterbe:
See comment above. Also, have you tested this on both python2 and python3?

kengruven:
Yes, it works and generates identical output with Python 2.7 and Python 3.4 for me.

peterbe:
Awesomeness! Would you mind rebasing it down into 1 commit? 

Also, it's annoying that the coveralls test isn't working.

peterbe:
Actually, when I run the tests myself, I see that there are only 2 lines missing test coverage in the whole repo, and that this
screenshot 2015-10-22 12 18 47

:)

PR #121, opened 2015-04-22
convert README.md to RST for PyPI

graingert:


peterbe:
But I like Markdown. It's much friendlier than ReST.
When will PyPI support both?

peterbe:
As you can see in the travis tests, you'll need to [this line](https://github.com/graingert/premailer/blob/fix-readme-on-pypi/setup.py#L16) too.

peterbe:
Would you mind rebasing and incorporating [the change that sneaked in](https://github.com/peterbe/premailer/commit/2034a2fb154fc0dbb87baff44329aa7db17bbbb5)?

ionelmc:
+1

PyPI doesn't support Markdown (there are no plans to support it AFAIK) and RST is quite good anyway.

peterbe:
Once again @graingert impresses me!

PR #123, opened 2015-05-04
Add support for URL without protocol (closes #119)

### LINKED ISSUE #119: Replacement of urls beginning with //
When using a link whose href starts with "//", Premailer should just add http or https depending on the base url it uses. Currently, it replaces it with the full host, so ``//another.host/`` becomes ``http://my.base.host/another.host/``

ewjoachim:
(See #119 for details)

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -417,6 +417,12 @@ class Premailer(object):
                         parent.attrib[attr].startswith('cid:')
                     ):
                         continue
+                    if parent.attrib[attr].startswith("//"):
+                        try:
+                            protocol = self.base_url.split("//")[0]
+                        except IndexError:
+                            protocol = "http"
+                        parent.attrib[attr] = protocol + parent.attrib[attr]
                     parent.attrib[attr] = urljoin(
                         self.base_url,
                         parent.attrib[attr].lstrip('/')
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -417,16 +417,18 @@ class Premailer(object):
                         parent.attrib[attr].startswith('cid:')
                     ):
                         continue
-                    if parent.attrib[attr].startswith("//"):
-                        try:
-                            protocol = self.base_url.split("//")[0]
-                        except IndexError:
-                            protocol = "http"
-                        parent.attrib[attr] = protocol + parent.attrib[attr]
-                    parent.attrib[attr] = urljoin(
+                    url = parent.attrib[attr]
+
+                    if not url.startswith("//"):
+                        url = url.lstrip('/')
+
+                    url = urljoin(
                         self.base_url,
-                        parent.attrib[attr].lstrip('/')
+                        url
                     )
+                    if url.startswith("//"):
+                        url = "http:" + url
+                    parent.attrib[attr] = url
 
         if hasattr(self.html, "getroottree"):
             return root
--- a/premailer/tests/test_premailer.py
+++ b/premailer/tests/test_premailer.py
@@ -2305,25 +2305,33 @@ sheet" type="text/css">
 
     def test_links_without_protocol(self):
         """If you the base URL is set to https://example.com and your html
-        contains <a href="//otherdomain.com/">...</a> then the URL to point to
+        contains <img src="//otherdomain.com/">... then the URL to point to
         is "https://otherdomain.com/" not "https://example.com/file.css"
         """
         html = """<html>
         <head>
         </head>
         <body>
-        <a href="//example.com">Link</a>
+        <img src="//example.com">
         </body>
         </html>"""
-
-        p = Premailer(html, base_url='https://www.peterbe.com')
-        result_html = p.transform()
 
         expect_html = """<html>
         <head>
         </head>
         <body>
-        <a href="https://example.com">Link</a>
+        <img src="{protocol}://example.com">
         </body>
         </html>"""
-        compare_html(expect_html, result_html)
+
+        p = Premailer(html, base_url='https://www.peterbe.com')
+        result_html = p.transform()
+        compare_html(expect_html.format(protocol="https"), result_html)
+
+        p = Premailer(html, base_url='http://www.peterbe.com')
+        result_html = p.transform()
+        compare_html(expect_html.format(protocol="http"), result_html)
+
+        p = Premailer(html, base_url='www.peterbe.com')
+        result_html = p.transform()
+        compare_html(expect_html.format(protocol="http"), result_html)
--- a/premailer/tests/test_premailer.py
+++ b/premailer/tests/test_premailer.py
@@ -2302,3 +2302,28 @@ sheet" type="text/css">
         result_html = p.transform()
 
         compare_html(expect_html, result_html)
+
+    def test_links_without_protocol(self):
+        """If you the base URL is set to https://example.com and your html
+        contains <a href="//otherdomain.com/">...</a> then the URL to point to
+        is "https://otherdomain.com/" not "https://example.com/file.css"
+        """
+        html = """<html>
+        <head>
+        </head>
+        <body>
+        <a href="//example.com">Link</a>
+        </body>
+        </html>"""
+
+        p = Premailer(html, base_url='https://www.peterbe.com')
+        result_html = p.transform()
+
+        expect_html = """<html>
+        <head>
+        </head>
+        <body>
+        <a href="https://example.com">Link</a>
+        </body>
+        </html>"""
+        compare_html(expect_html, result_html)


premailer/premailer.py
```diff
@@ -417,6 +417,12 @@ def transform(self, pretty_print=True, **kwargs):
                         parent.attrib[attr].startswith('cid:')
                     ):
                         continue
+                    if parent.attrib[attr].startswith("//"):
```
graingert: This is already supported if you set
```transform('...', base_url='http://www.peterbe.com/')```

https://github.com/peterbe/premailer#turning-relative-urls-into-absolute-urls

**Comment by graingert** [comment]
This is already supported if you set
```transform('...', base_url='http://www.peterbe.com/')```

https://github.com/peterbe/premailer#turning-relative-urls-into-absolute-urls

File: premailer/premailer.py [hunk 2]
```diff
@@ -417,6 +417,12 @@ def transform(self, pretty_print=True, **kwargs):
...
+                        try:
+                            protocol = self.base_url.split("//")[0]
+                        except IndexError:
+                            protocol = "http"
```
graingert: http is deprecated. Use https.

premailer/premailer.py
graingert: You can just use urljoin here
```
>>> from urlparse import urljoin
>>> urljoin('https://example.com', '//google.com/foo.html')
'https://google.com/foo.html'
>>> urljoin('https://google.com', 'http://google.com/foo.html')
''http://google.com/foo.html'
```

graingert:
This looks like it's a bug in lxml.

`.make_links_absolute` should add the scheme in this case

premailer/premailer.py
peterbe: +1 on using `urljoin()`

premailer/premailer.py
peterbe: +1

premailer/tests/test_premailer.py
```diff
@@ -2302,3 +2302,28 @@ def test_turnoff_cache_works_as_expected(self):
...
+    def test_links_without_protocol(self):
+        """If you the base URL is set to https://example.com and your html
+        contains <a href="//otherdomain.com/">...</a> then the URL to point to
+        is "https://otherdomain.com/" not "https://example.com/file.css"
+        """
+        html = """<html>
+        <head>
+        </head>
+        <body>
+        <a href="//example.com">Link</a>
```
peterbe: Hmm... It makes me wonder. Why ever allow or bother with links like this?
The `(href|src)="//...` trick is "reserved" for inclusion of static assets (`script`, `link`, `img`, etc). That's because you'll want to load multiple requests under one page view and when you do that the browser dictates that you'll want them also loaded on HTTPS if you're currently on and HTTPS page. 

However, if you have a link that's not going to be part of the page view. It's for something entirely different later. 

If you want to make a link I think the right thing to do is something like `<a href="otherpage.html">` or `<a href="/relative/abs.html">` (or `<a href="http://specificdomain.com/index.html">` of course). 

I'm sorry if I'm making your life hard here :) 
But I'm not convinced premailer should get itself involved in (bad) HTML choices. Perhaps I'm wrong about using `//` in hyperlinks and I'm more than happy to be educated and convinced.

graingert: I don't think premailer should be hacking a bug fix around lxml.

Can we move this issue to their tracker? Maybe they have a better
understanding why lxml doesn't do this already?

Authors should really just use https:// wherever they use protocol relative
URLs.

premailer/premailer.py
ewjoachim: Hmm, I'm not sure... If we don't touch it, we're letting mail client handle "//" links and they don't really do it well (as far as I can tell, at least OSX Mail and Thunderbird don't like it. Gmail does it well but it's in a browser so it's cheating...). We're potentially talking about the combination of :

 - Using ```"mysite.com"``` as base url
 - Having ``````

that would render to `````` which mail client would not understand, I'm not convinced it's the best choice... Or maybe I overlooked something ?

(Note that if the base url itself is defined as ```"//mysite.com"``` that would make links defined the same way with my implementation (but then you get what you asked for).)

Maybe a solution would be to just remove ```//``` so we would get simply `````` that mail clients will probably interpret as http

graingert: Where are these '//' URLs coming from?

premailer/premailer.py
ewjoachim: Oh, didn't know that. Seems nice...

premailer/premailer.py
ewjoachim: Mainly from the fact I'm using the same django template for both a web view and a mail, that I'm using premailer on both cases (because I want te be able, as far as possible, to detect a possible problem in the premailer transformation of my e-mail before sending), that the web template has some js in it and that the js is linked with //. I was just surprised when my // link did not work, and submitted the issue. If finally it's not an issue, we can reject the PR and call it a day. I'll put https links instead of // (or not premail the template for the web) and I'll be ok.

I don't know if could be useful, but I've seen a situation where a website used both http and https and most static files (including images) (stored on another domain) were linked using // to not mix http (which had a lot of traffic) and https (which didn't). But then I can't say it's a frequent setup, so it may not be worth the effort...

premailer/premailer.py
```diff
@@ -417,6 +417,12 @@ def transform(self, pretty_print=True, **kwargs):
                         parent.attrib[attr].startswith('cid:')
                     ):
                         continue
+                    if parent.attrib[attr].startswith("//"):
+                        try:
+                            protocol = self.base_url.split("//")[0]
```
ewjoachim: Yup, it's missing a ```:``` and missing a test for this path

premailer/tests/test_premailer.py
ewjoachim: Hmm, but we're not calling ```make_links_absolute ``` as far as I can tell, so I don't really get it...?

And sometimes (probably less and less frequently but it still happens), http is the right choice for performance...

ewjoachim: @peterbe : Yes, I've been using a ```<a href>``` to demonstrate the bug where I maybe should have used a ``````. I'll modify the test but it's the same code.

graingert: It doesn't matter about performance. HTTP is deprecated.

ewjoachim: well, as long as "not providing a protocol" will mean "using http" for all major browsers, I'm afraid it's a little bit utopic...

premailer/premailer.py
```diff
@@ -417,10 +417,18 @@ def transform(self, pretty_print=True, **kwargs):
                         parent.attrib[attr].startswith('cid:')
                     ):
                         continue
-                    parent.attrib[attr] = urljoin(
+                    url = parent.attrib[attr]
+
+                    if not url.startswith("//"):
+                        url = url.lstrip('/')
```
peterbe: I struggle to parse this. So, if `url` is something like `/path/to.css` you're going to change that to `path/to.css`. That is not necessary because `urljoin()` is really clever:
```python
>>> urljoin('https://www.peterbe.com/', '/path/to.css')
'https://www.peterbe.com/path/to.css'
>>> urljoin('https://www.peterbe.com/', 'path/to.css')
'https://www.peterbe.com/path/to.css'
>>> urljoin('https://www.peterbe.com', '/path/to.css')
'https://www.peterbe.com/path/to.css'
>>> urljoin('https://www.peterbe.com', 'path/to.css')
'https://www.peterbe.com/path/to.css'
```

Best of all, `urljoin()` understands `//` URLs too. E.g. 
```
>>> urljoin('https://www.peterbe.com', '//domain.com/path/to.css')
'https://domain.com/path/to.css'
>>> urljoin('http://www.peterbe.com', '//domain.com/path/to.css')
'http://domain.com/path/to.css'
```
which is ideal.

peterbe:
Hmm... This is getting interesting. 

I was eager to say we should use `urljoin()` everywhere for everything but I see there's a test that allows you to set the `base_url` to something like `https://kungfupeople.com/base` and then it should merge that with `/images/foo.jpg` and become `https://kungfupeople.com/base/images/foo.jpg`. 

I'm starting to think that's weird. Why would you say `` if you actually mean ``. 

All this `url.lstrip('/')` stuff makes me nervous. I'm tempted to delete it all and just use `urljoin()` and re-write the tests that break. If we do that we might want to bump the version number a bit. 

What do you think @graingert? Should we be bold and scrap the hacks?

By the way @ewjoachim, what the above proposal outlines is a solution to your original problem. Basically, if you have:
```html
<img src="//cdn.com/foo.jpg">
``` 
and the `base_url` set to `https://www.example.com` the final output becomes:
```html
<img src="https://cdn.com/foo.jpg">
```

peterbe:
Also, I propose we add this:
```diff
         # URLs
         #
         if self.base_url:
-            if not self.base_url.endswith('/'):
-                self.base_url += '/'
+            if '://' not in self.base_url:
+                raise ValueError('Base URL must start have a protocol')
+            # if not self.base_url.endswith('/'):
+            #     self.base_url += '/'
             for attr in ('href', 'src'):
                 for item in page.xpath("//@%s" % attr):
                     parent = item.getparent()
```
Basically, we disallow people to set `Premailer(html, base_url='example.com')`. If this was 2010 I would be content with defaulting the user over to `http://` but it's 2015 and `https://` is all the new rage. However, not everyone has made the switch yet so it's just too hard to make up a default. Putting that responsibility on the user seems a better idea for a lib.

peterbe:
With @ewjoachim's blessing, we're attacking this in [PR #124](https://github.com/peterbe/premailer/pull/124) instead.

PR #124, opened 2015-05-08
always urljoin - fixes #119

### LINKED ISSUE #119: Replacement of urls beginning with //

peterbe:
What do you think @graingert and @ewjoachim ?

* This solves the problem of `` picking up the protocol from the `base_url`.
* Disallows ambiguous `base_url` values like `www.example.com` because you don't want to allow this to be in the output `` because when viewed in a web page it'll load, for example, `https://mail.google.com/www.example.com/bar.png`
* If you set the `base_url` to `https://example.com/folder/` and you have `` the output becomes `` instead.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -17,7 +17,7 @@ if sys.version_info >= (3,):  # pragma: no cover
     # As in, Python 3
     from io import StringIO
     from urllib.request import urlopen
-    from urllib.parse import urljoin
+    from urllib.parse import urljoin, urlparse
     STR_TYPE = str
 else:  # Python 2
     try:
@@ -26,7 +26,7 @@ else:  # Python 2
         from StringIO import StringIO
         StringIO = StringIO  # shut up pyflakes
     from urllib2 import urlopen
-    from urlparse import urljoin
+    from urlparse import urljoin, urlparse
     STR_TYPE = basestring
 
 import cssutils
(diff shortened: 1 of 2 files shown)


graingert:
Use
 
`if urlparse.urlparse(base_url).scheme:`

premailer/premailer.py
```diff
@@ -402,25 +402,23 @@ def transform(self, pretty_print=True, **kwargs):
         # URLs
         #
         if self.base_url:
-            if not self.base_url.endswith('/'):
-                self.base_url += '/'
+            if '://' not in self.base_url:
+                raise ValueError('Base URL must start have a protocol')
```
graingert: Fix the grammar. Also URLs have schemes not protocols.

peterbe:
@graingert done.

peterbe:
@graingert I force-pushed a new commit message. Just for you :)

graingert:
Bling

peterbe:
So, @graingert you approve?

What do you think @ewjoachim ?

ewjoachim:
Nothing to add, I'm ok.

premailer/premailer.py
```diff
@@ -402,25 +402,23 @@ def transform(self, pretty_print=True, **kwargs):
         # URLs
         #
         if self.base_url:
-            if not self.base_url.endswith('/'):
-                self.base_url += '/'
+            if not urlparse(self.base_url).scheme:
+                raise ValueError('Base URL must start have a protocol')
```
ewjoachim: Erm, I think the phrasing of the Value Error is still wrong... And @graingert suggested using "scheme" instead of protocol (which would also apply to the name of the test)

premailer/tests/test_premailer.py
```diff
@@ -505,7 +503,7 @@ def test_base_url_with_path(self):
         </body>
         </html>'''
 
-        p = Premailer(html, base_url='http://kungfupeople.com/base',
+        p = Premailer(html, base_url='http://kungfupeople.com/base/',
```
ewjoachim: (Is there still a test for what happens if there's no trailing slash ?)

peterbe: I don't think there is. But the result should be identical to what `urljoin` does.

premailer/premailer.py
peterbe: Ha! My bad. I thought he pointed out that I had put "schema" (not "scheme") in the commit message. But no, I had put "schema" in the commit message and "protocol" there in the ValueError message. Fixed now.

PR #126, opened 2015-05-12
Some improvements for inlining css in a "fragment"

ionelmc:
I have an usecase where I don't have any body/html tags, therefore no head.

There are two changes here:
* the `add_leftover_css` option to disable injecting letfovers into head
* allow not having/finding a head at all - is there a better way to handle this?

Also, what's a good example test to look at (for having tests for these changes)?

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -113,7 +113,8 @@ class Premailer(object):
                  disable_validation=False,
                  cache_css_parsing=True,
                  cssutils_logging_handler=None,
-                 cssutils_logging_level=None):
+                 cssutils_logging_level=None,
+                 disable_leftover_css=False):
         self.html = html
         self.base_url = base_url
         self.preserve_internal_links = preserve_internal_links
@@ -139,6 +140,7 @@ class Premailer(object):
         self.disable_basic_attributes = disable_basic_attributes
         self.disable_validation = disable_validation
         self.cache_css_parsing = cache_css_parsing
+        self.disable_leftover_css = disable_leftover_css
 
         if cssutils_logging_handler:
             cssutils.log.addHandler(cssutils_logging_handler)
@@ -242,7 +244,6 @@ class Premailer(object):
                         len(rules)  # this is the rule's index number
                     )
                     rules.append((specificity, selector, bulk))
-
         return rules, leftover
 
     def transform(self, pretty_print=True, **kwargs):
@@ -271,7 +272,10 @@ class Premailer(object):
 
         assert page is not None
 
-        head = get_or_create_head(tree)
+        if self.disable_leftover_css:
+            head = None
+        else:
+            head = get_or_create_head(tree)
         #
         # style selectors
         #
@@ -553,7 +557,7 @@ class Premailer(object):
         """
         these_rules, these_leftover = self._parse_style_rules(css_text, index)
         rules.extend(these_rules)
-        if these_leftover or self.keep_style_tags:
+        if head is not None and (these_leftover or self.keep_style_tags):
             style = etree.Element('style')
             style.attrib['type'] = 'text/css'
             if self.keep_style_tags:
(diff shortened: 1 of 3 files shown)


premailer/premailer.py
```diff
@@ -111,7 +114,8 @@ def __init__(self, html, base_url=None,
                  base_path=None,
                  disable_basic_attributes=None,
                  disable_validation=False,
-                 cache_css_parsing=True):
+                 cache_css_parsing=True,
+                 add_leftover_css=True):
```
peterbe: Can you reverse the meaning of this parameter. It's generally a good idea to have "negative" defaults. E.g. instead you can make it `disable_leftover_css=False`.

ionelmc: Fine. How about the questions I had in the PR?

peterbe:
Pardon the delay. I'm sorry for not being able to give a better review. I do not doubt your need that you have. Let's try to fix it. 

I like the idea of making it an option to disable that the left over CSS gets inserted in the `<head>`. But I can't see why you need to also change the `get_or_create_head` method. It would be better if you could just entirely avoid it with one simple setting.

ionelmc:
Well yeah, the `get_or_create_head` change is a bit unrelated to the `leftover_css`. However, it's a necessary change as a missing head/body doesn't need to be a fatal error. Right?

peterbe:
Let's just solve one problem at a time. Can you chop away the change related to `get_or_create_head` for now? 

Actually, I think that if you want to put the left over CSS back in you need to have a `<head>` and a `<body>`. If someone has a problem with they should use your fancy new option :)

ionelmc:
Fine. About them tests, where's are the tests for the normal leftover head injection behavior?

peterbe:
> Fine. About them tests, where's are the tests for the normal leftover head injection behavior?

Good question :)

What I would do is put in a `raise Exception` near the area where you expect to work and then run `nosetests premailer` to see which tests cover that area (as they'd break there). Then I would simply scan the test methods in `test_premailer.py` till you see something that is somewhat matching.

premailer/premailer.py
```diff
@@ -235,8 +237,10 @@ def join_css_properties(properties):
                         len(rules)  # this is the rule's index number
                     )
                     rules.append((specificity, selector, bulk))
-
-        return rules, leftover
+        if not self.disable_leftover_css:
+            return rules, leftover
+        else:
+            return rules, None
```
peterbe: This change isn't necessary. Let it return `rules, leftover` here. 
Then [here](https://github.com/ionelmc/premailer/blob/master/premailer/premailer.py#L315) instead you change it to 
```python
if (
    (these_leftover or self.keep_style_tags) and
    not self.disable_leftover_css
):
```
At least I think that would work, but some good new tests would be best to prove it.

premailer/premailer.py
```diff
@@ -264,7 +268,10 @@ def transform(self, pretty_print=True, **kwargs):
 
         assert page is not None
 
-        head = get_or_create_head(tree)
+        if self.disable_leftover_css:
+            head = None
+        else:
+            head = get_or_create_head(tree)
```
peterbe: This might not be enough. You might also need to add `and not self.disable_leftover_css` [here](https://github.com/ionelmc/premailer/blob/master/premailer/premailer.py#L336) and [here](https://github.com/ionelmc/premailer/blob/master/premailer/premailer.py#L343) too.

ionelmc:
Alright, squashed the commits, added a test and fixed the checks for the head.

tox.ini
```diff
@@ -7,6 +7,7 @@
 envlist = flake8, py26, py27, py32, py33, py34, pypy
 
 [testenv]
+passenv = *
```
peterbe: What is this?

ionelmc: Your suite fails on tox-2.0. See: https://testrun.org/tox/latest/config.html#confval-passenv=SPACE-SEPARATED-GLOBNAMES

ionelmc: At least on Windows it fails.

PR #134, opened 2015-08-14
Avoid skipping links with media="all"

jvrsantacruz:
Links with `media="all"` target all devices [as the spec says](http://www.w3.org/TR/CSS2/media.html#media-types), but premailer only processes links with `media="screen"`.

For example, this `link` will be ignored, and it shouldn't:

```
        <link rel="stylesheet" media="all" type="text/css" href="styles.css">
```

Changed check to process media types `"all"` and `"screen"`.


peterbe:
Welcome to the project! 
I'll make a new release for this.

jvrsantacruz:
Thanks! So fast! :)

PR #135, opened 2015-08-27
Add support for adding align tags on images from css float

Tigge:
Outlook (both on desktop and the web version) are notably bad at parsing CSS. One improvements if you want to use float on images is to use the align attribute instead of css float. Is this something that could be interesting to add into premailer? If so, then I'll clean up this pull request a bit, fix the code, and do some more testing.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -114,7 +114,8 @@ class Premailer(object):
                  cache_css_parsing=True,
                  cssutils_logging_handler=None,
                  cssutils_logging_level=None,
-                 disable_leftover_css=False):
+                 disable_leftover_css=False,
+                 align_floating_images=True):
         self.html = html
         self.base_url = base_url
         self.preserve_internal_links = preserve_internal_links
@@ -141,6 +142,7 @@ class Premailer(object):
         self.disable_validation = disable_validation
         self.cache_css_parsing = cache_css_parsing
         self.disable_leftover_css = disable_leftover_css
+        self.align_floating_images = align_floating_images
 
         if cssutils_logging_handler:
             cssutils.log.addHandler(cssutils_logging_handler)
@@ -408,6 +410,17 @@ class Premailer(object):
             for item in page.xpath('//@class'):
                 parent = item.getparent()
                 del parent.attrib['class']
+
+        # Add align attributes to images if they have a CSS float value of
+        # right or left. Outlook (both on desktop and on the web) are bad at
+        # understanding floats, but they do understand the HTML align attrib.
+        if self.align_floating_images:
+            for item in page.xpath('//img[@style]'):
+                image_css = cssutils.parseStyle(item.attrib['style'])
+                if image_css.float == 'right':
+                    item.attrib['align'] = 'right'
+                elif image_css.float == 'left':
+                    item.attrib['align'] = 'left'
 
         #
         # URLs
--- a/premailer/tests/test_premailer.py
+++ b/premailer/tests/test_premailer.py
@@ -2419,3 +2419,37 @@ sheet" type="text/css">
         # Because you can't set a base_url without a full protocol
         p = Premailer(html, base_url='www.peterbe.com')
         assert_raises(ValueError, p.transform)
+
+    def test_align_float_images(self):
+
+        html = """<html>
+        <head>
+        <title>Title</title>
+        <style>
+        .floatright {
+            float: right;
+        }
+        </style>
+        </head>
+        <body>
+        <p><img src="/images/left.jpg" style="float: left"> text
+           <img src="/images/right.png" class="floatright"> text
+           <img src="/images/nofloat.gif"> text
+        </body>
+        </html>"""
+
+        expect_html = """<html>
+        <head>
+        <title>Title</title>
+        </head>
+        <body>
+        <p><img src="/images/left.jpg" style="float: left" align="left"> text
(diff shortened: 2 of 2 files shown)


premailer/premailer.py
```diff
@@ -409,6 +411,15 @@ def transform(self, pretty_print=True, **kwargs):
...
 
+        if self.image_align:
+            for item in page.xpath("//img"):
+                image_css = cssutils.parseStyle(item.attrib['style'])
+                if image_css.float == "right":
+                    item.attrib["align"] = "right"
+                elif image_css.float == "left":
+                    item.attrib["align"] = "left"
+
+
```
peterbe: This excess newlines is what's breaking the travis tests.

premailer/premailer.py
```diff
@@ -114,7 +114,8 @@ def __init__(self, html, base_url=None,
                  cache_css_parsing=True,
                  cssutils_logging_handler=None,
                  cssutils_logging_level=None,
-                 disable_leftover_css=False):
+                 disable_leftover_css=False,
+                 image_align=False):
```
peterbe: The word "image_align" is a bit vague. What about "align_images" or "align_floating_images"?

peterbe:
I think we should do it but it needs a bit more paperwork. 
To start, the wonderful description you put here in the PR comment you should "copy" in as a comment in the code. 
We'll also need a unit test that uses this feature.

Tigge:
Added a unit test and cleaned up the code a bit - also added a small comment to the code.

premailer/premailer.py
```diff
@@ -409,6 +411,18 @@ def transform(self, pretty_print=True, **kwargs):
                 parent = item.getparent()
                 del parent.attrib['class']
 
+        # Add align attributes to images if they have a CSS float value of
+        # right or left. Outlook (both on desktop and the web version) are
+        # bad at parsing CSS (like floats) but they do understand the HTML
+        # align attribute.
+        if self.align_floating_images:
+            for item in page.xpath("//img"):
+                image_css = cssutils.parseStyle(item.attrib.get('style', ''))
```
peterbe: This we might want to optimize a bit. 
For every single image that does NOT have a `style` attribute (not sure that there's that many) we have to go through the parse and the if statements below. How about something like this:
```python
for item in page.xpath('//img[@style]'):
    image_css = cssutils.parseStyle(item.attrib['style'])
    if ...
```

premailer/premailer.py
```diff
@@ -409,6 +411,18 @@ def transform(self, pretty_print=True, **kwargs):
                 parent = item.getparent()
                 del parent.attrib['class']
 
+        # Add align attributes to images if they have a CSS float value of
+        # right or left. Outlook (both on desktop and the web version) are
+        # bad at parsing CSS (like floats) but they do understand the HTML
+        # align attribute.
+        if self.align_floating_images:
+            for item in page.xpath("//img"):
```
peterbe: single quotes please. 
I know it doesn't make any run-time difference but good code tries to be consistent.

premailer/premailer.py
```diff
@@ -409,6 +411,18 @@ def transform(self, pretty_print=True, **kwargs):
...
+                image_css = cssutils.parseStyle(item.attrib.get('style', ''))
+                if image_css.float == "right":
+                    item.attrib["align"] = "right"
+                elif image_css.float == "left":
+                    item.attrib["align"] = "left"
```
peterbe: single quotes here too please

premailer/premailer.py
```diff
@@ -409,6 +411,18 @@ def transform(self, pretty_print=True, **kwargs):
                 parent = item.getparent()
                 del parent.attrib['class']
 
+        # Add align attributes to images if they have a CSS float value of
+        # right or left. Outlook (both on desktop and the web version) are
+        # bad at parsing CSS (like floats) but they do understand the HTML
```
peterbe: "bad at parsing CSS (like floats)"
That's probably true but that expression is too broad for this particular case. The problem is that they don't understand floats. 
How about phrasing it like "...are bad at understanding CSS floats, but they do under the HTML align attribute." ?

premailer/tests/test_premailer.py
```diff
@@ -2419,3 +2419,37 @@ def test_links_without_protocol(self):
...
+        <p><img src="/images/left.jpg" style="float: left" align="left"> text
+           <img src="/images/right.png" style="float:right" align="right"> text
+           <img src="/images/nofloat.gif"> text
+        </p>
+        </body>
+        </html>"""
+
+        p = Premailer(html, align_floating_images=True)
+        result_html = p.transform()
+        compare_html(expect_html, result_html)
```
peterbe: Great test!

premailer/premailer.py
```diff
@@ -114,7 +114,8 @@ def __init__(self, html, base_url=None,
                  cache_css_parsing=True,
                  cssutils_logging_handler=None,
                  cssutils_logging_level=None,
-                 disable_leftover_css=False):
+                 disable_leftover_css=False,
+                 align_floating_images=False):
```
peterbe: I'm not sure this should be off by default.
It seems like such a useful feature. If it's off by default only people with deep insight into Outlook will benefit and others will miss out. 
Let's just change it to by True by default.

peterbe:
screenshot 2015-09-08 08 29 11

It works!

Tigge:
Yay!

PR #136, opened 2015-09-04
add `unset_removes_attribute` option

elidickinson:
The purpose of this option is to give HTML email authors an easy way to reduce generated file size.

This option instructs premailer skip the inlining of CSS attributes that are set to `unset` because `unset` has effectively the same meaning as not inlining the property ([more](https://developer.mozilla.org/en-US/docs/Web/CSS/unset)). So a set of CSS rules that would previously have resulted in: `<p style="font-size: 14px; font-family: unset">` will instead result in `<p style="font-size: 14px">`, which should render exactly the same in all clients, but saves bytes. 

Support for `unset` is [not unviersal](http://caniuse.com/#feat=css-unset-value), but I don't this is a problem here. In my intended use of this feature, you would never actually see `unset` in the output from premailer. You could technically accomplish a similar thing by using a made-up property like `premailer-dont-inline-this` but it's a nice bonus that `unset` is a semi-official property with roughly the same meaning and is supported natively by Chrome and Firefox.

I created this as an off-by-default option to be conservative, but I actually think it could be merged into the core. Almost nobody is using `unset` in HTML emails, so for them this has no impact on the output. For people who are using it, this should only improve the output by removing a "meaningless" attribute that could confuse email clients.

By adding a couple of simple rules to our style sheet that look kinda like this:
```css
html body td.expander {
  /* expanders never contain text so font formatting is irrelevant */
  color: unset;
  font-family: unset;
  font-weight: unset;
  font-size: unset;
}
```
we took a 130kb reference newsletter (based on Zurb Ink) and shrunk it to 95kb.

### DIFF
--- a/premailer/merge_style.py
+++ b/premailer/merge_style.py
@@ -19,7 +19,12 @@ def csstext_to_pairs(csstext):
 csstext_to_pairs._lock = threading.RLock()
 
 
-def merge_styles(inline_style, new_styles, classes):
+def merge_styles(
+        inline_style,
+        new_styles,
+        classes,
+        remove_unset_properties=False
+):
     """
         This will merge all new styles where the order is important
         The last one will override the first
@@ -29,9 +34,12 @@ def merge_styles(inline_style, new_styles, classes):
 
         Args:
             inline_style(str): the old inline style of the element if there
-            is one new_styles: a list of new styles, each element should be
-            a list of tuple classes: a list of classes which maps
-            new_styles, important!
+                is one
+            new_styles: a list of new styles, each element should be
+                a list of tuple
+            classes: a list of classes which maps new_styles, important!
+            remove_unset_properties(bool): Allow us to remove certain CSS
+                properties with rules that set their value to 'unset'
 
         Returns:
             str: the final style
@@ -55,6 +63,13 @@ def merge_styles(inline_style, new_styles, classes):
     normal_styles = []
     pseudo_styles = []
     for pseudoclass, kv in styles.items():
+        if remove_unset_properties:
+            # Remove rules that we were going to have value 'unset' because
+            # they effectively are the same as not saying anything about the
+            # property when inlined
+            kv = dict(
+                (k, v) for (k, v) in kv.items() if not v.lower() == 'unset'
+            )
         if not kv:
             continue
         if pseudoclass:
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -115,7 +115,8 @@ class Premailer(object):
                  cssutils_logging_handler=None,
                  cssutils_logging_level=None,
                  disable_leftover_css=False,
-                 align_floating_images=True):
+                 align_floating_images=True,
+                 remove_unset_properties=True):
         self.html = html
         self.base_url = base_url
         self.preserve_internal_links = preserve_internal_links
@@ -143,6 +144,7 @@ class Premailer(object):
         self.cache_css_parsing = cache_css_parsing
         self.disable_leftover_css = disable_leftover_css
         self.align_floating_images = align_floating_images
+        self.remove_unset_properties = remove_unset_properties
 
         if cssutils_logging_handler:
             cssutils.log.addHandler(cssutils_logging_handler)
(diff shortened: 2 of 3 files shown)


premailer/merge_style.py
```diff
@@ -31,7 +31,8 @@ def merge_styles(inline_style, new_styles, classes):
             inline_style(str): the old inline style of the element if there
             is one new_styles: a list of new styles, each element should be
             a list of tuple classes: a list of classes which maps
-            new_styles, important!
+            new_styles, important! unset_removes_attribute: allow us to remove
+            certain CSS attributes with rules that set their value to 'unset'
```
peterbe: It was already a strange sentence. But Did you perhaps mean to prefix the `unset_removes_attribute: allow us to ...` with some newlines?

premailer/premailer.py
```diff
@@ -395,8 +397,11 @@ def transform(self, pretty_print=True, **kwargs):
         # and a long list of elements
         for _, element in elements.items():
             final_style = merge_styles(element['item'].attrib.get('style', ''),
-                                       element['style'], element['classes'])
-            element['item'].attrib['style'] = final_style
+                                       element['style'], element['classes'],
+                                       unset_removes_attribute=self.unset_removes_attribute)
```
peterbe: Can you please rewrite these lines to be:
```python
final_style = merge_styles(
    element['item'].attrib.get('style', ''),
    element['style'], 
    element['classes'],
    unset_removes_attribute=self.unset_removes_attribute,
)
```
That kind of formatting has the advantage that it's indentation for line 2 and 3 isn't dependent of the length of the variable it assigns to (e.g. `final_style`).
Also, note the trailing comma on the last parameter.

premailer/tests/test_premailer.py
```diff
@@ -2419,3 +2429,37 @@ def test_links_without_protocol(self):
...
+        </head>
+        <body>
+        <div style="color:green"><span></span></div>
+        </body>
+        </html>"""
+
+        p = Premailer(html, unset_removes_attribute=True)
+        self.assertTrue(p.unset_removes_attribute)
+        result_html = p.transform()
+        print result_html
```
peterbe: debugging left-overs presumably :)

peterbe:
GREAT WORK! I like it!
As you can see you have some Travis errors to deal with. They should be relatively easy to address. Let me know if they give you grief. 

Two things bother me...

One, if someone wants to use this, they have to first deliberately set `unset` values on their selector values **and** pass in `unset_removes_attribute=TRUE` for it to work. I think it should be on by default. After all, the purpose of Premailer is to make life easier for people. 

Two, the name "attribute" doesn't feel right. If you have `A { B: C; }` the `C` part is called a "value" and `B` is the "property". Together I don't know if they have name. Perhaps "Property value pair". So, perhaps "unset_removes_property" is more accurate. 

Having said that, if you look at [the other list of options](https://github.com/peterbe/premailer/blob/c642b55a36c55e0da9a0b712626369c50a8f382e/premailer/premailer.py#L101-L117) you'll notice there's a pattern of verbs. "preserve", "keep", "disable", "remove", "strip". 
Which makes me think this ought to be "remove_unset_properties". What do you think?

peterbe:
Note that I recently merged https://github.com/peterbe/premailer/pull/135 which I think will cause a merge conflict with your code. If you need help rebasing your branch, let me know.

dbarbar:
My original attempt at a rebase got messy, so I ended up rebasing against master and squashing these commits and force pushing this branch. Sorry for rewriting history. Hopefully this'll make review easier though.

I also did the following changes:
1. Fix long lines and other pep8 things.
2. Remove the print statement from the new test.
3. Rename the setting to "remove_unset_properties".
4. Default the setting to True.

Looks like Travis is happy with the changes. Let me know if this needs more work. Thanks!

peterbe:
Strange that the author changed. Either way, it's been merged now. I like it! I'll prep a new release for it.

dbarbar:
Thanks! The author change isn't too strange, I'm a colleague of @elidickinson here at Industry Dive, so I got to clean things up, after making them messier.

peterbe:
Industry Dive? Interesting. Would you be interested in sponsoring premailer?

This feature is now fully incorporated on http://premailer.io/ Please try it out. I did. It seems to work as expected.

PR #141, opened 2015-10-07
Suppress base_url via attribute

### LINKED ISSUE #140: Phone number URLs are broken
Valid `tel:` urls are prepended with base_url when they shouldn't be. This is somehow related to #8.

May I suggest adding an attribute that disables rewriting for a specific `<a>` tag? Maybe something like this:

``` html
<a href="tel:12345" data-premailer="ignore">Link</a>
```

viernullvier:
Don't prepend base_url to URLs if the tag's attribute 'data-premailer' is set to 'ignore'. Helps with / fixes #140.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -441,6 +441,8 @@ class Premailer(object):
                 for item in page.xpath("//@%s" % attr):
                     parent = item.getparent()
                     url = parent.attrib[attr]
+                    if parent.attrib.pop(self.attribute_name, None) == 'ignore':
+                        continue
                     if (
                         attr == 'href' and self.preserve_internal_links and
                         url.startswith('#')


peterbe:
You have me convinced in the Issue that we need to tackle this with an option. However, the PR needs some work. I don't like that the `data-premailer="ignore"` attribute is still there after premailer has transformed. Also, if someone reads `data-premailer="ignore"` they might think that the **whole tag** should be ignored. E.g. if you have 
```
<style>a { color: red }</style>
<a href="tel:+1234" data-premailer="ignore">Call</a>
<a href="/">Home page</a>
```
...and giving someone sets `base_url='http://example.com'` to the transform, you might expect this output:
```
<a href="tel:+1234">Call</a>
<a href="http://example.com/" style="color:red">Home page</a>
```
which is NOT want we want. We just want the `href` attribute to be ignored. 

Do you have some other ideas about how to annotate and tell premailer to "not touch" an attribute value?

peterbe:
How about this for  a functional idea:
```python
def my_replacer(value, base_url):
      if 'tel:' in value:
         # leave it as is 
         return value
     return urlparse.urljoin(base_url, value)
premailer.transform(html, base_url='http://example.com', base_url_replacer=my_replacer)
```

peterbe:
So the problem is that it becomes `<a href="http://mybaseurl/tel:+1234">Call</a>`, right?
Can't we write a regex that checks the content of the `href` tag before it does the "insertion" of the base_url?

PR #142, opened 2015-10-14
Support non ASCII characters on output

jhutchins:
Without the encode step running in Python 2.7.10 (and probably other Python 2 versions) using the ```-o``` flag or redirecting the output with the ```>``` operator lead to a stack trace like

```
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/jhutchins/git/api/templates/env/lib/python2.7/site-packages/premailer/__main__.py", line 142, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Users/jhutchins/git/api/templates/env/lib/python2.7/site-packages/premailer/__main__.py", line 137, in main
    options.outfile.write(p.transform(pretty_print=options.pretty))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 20719: ordinal not in range(128)
```

### DIFF
--- a/premailer/__main__.py
+++ b/premailer/__main__.py
@@ -134,7 +134,12 @@ def main(args):
         disable_basic_attributes=options.disable_basic_attributes,
         disable_validation=options.disable_validation
     )
-    options.outfile.write(p.transform(pretty_print=options.pretty))
+    
+    output = p.transform(pretty_print=options.pretty)
+    if hasattr(output, 'encode'):  # Python 2 compatability
+        output = output.encode('utf-8')
+    options.outfile.write(output)
+    
     return 0
 
 


peterbe:
I'm not sure what's going on and will probably not have the head space to debug it with you but I'd suggest you start with writing a test case that demonstrates the current problem.

peterbe:
You're not alone :)
https://github.com/peterbe/premailer/issues/152

Would you mind taking a look at the travis errors and see if you can fix this PR?

mgilson:
A very simple example that reproduces this problem is the following:

```html
<style>
.foo {
    display: inline-block;
}
</style>

<html>
<head>
</head>
<body>
  <div class="foo">&bull;l</div>
</body>
</html>
```

Now if you run premailer from the commandline:

```
$ python -m premailer -f ~/test.html -o foo.html
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Library/Python/2.7/site-packages/premailer/__main__.py", line 142, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Library/Python/2.7/site-packages/premailer/__main__.py", line 137, in main
    options.outfile.write(p.transform(pretty_print=options.pretty))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2022' in position 74: ordinal not in range(128)
```

mgilson:
FWIW, if I use the ascii encoding, all seems to be right with the world...

```python
html="""
<html>
<head>
<style>
.foo {
    display: inline-block;
}
</style>
</head>
<body>
  <div class="foo">&bull;</div>
</body>
</html>
"""
import premailer

p = premailer.Premailer(html)
encoding = 'ascii'
# encoding = 'utf-8'  # This is broken
print p.transform(encoding='ascii')
```

This turns my `&bull;` into `&#8226;`, but that's no real problem since it's still valid HTML and I'm not planning on modifying the HTML with inlined CSS. . . :-)

So this might be a viable workaround for a number of users.

peterbe:
Try upgrading to 3.0.1 and use the new `--encoding` option on the command line.

PR #144, opened 2015-10-20
Add an option to capitalize margin and float properties

JocelynDelalande:
The option is disabled by default, that weird idea comes from a weird
outlook.com behaviour to handle margin and float properties only if the
first letter is uppercase.

See https://www.emailonacid.com/blog/article/email-development/outlook.com-does-support-margins

Feel free to give me some leads/hints if my approach is not good, but regexps is (sadly) the better I found that didn't required to modify *cssutils* and folks.

### DIFF
--- a/premailer/__main__.py
+++ b/premailer/__main__.py
@@ -65,6 +65,12 @@ def main(args):
     )
 
     parser.add_argument(
+        "--capitalize-float-margin", default=False,
+        help="Capitalize float and margin properties for outlook.com compat.",
+        action="store_true", dest="capitalize_float_margin"
+    )
+
+    parser.add_argument(
         "--strip-important", default=False,
         help="Remove '!important' for all css declarations.",
         action="store_true", dest="strip_important"
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -85,8 +85,27 @@ def _cache_parse_css_string(css_body, validate=True):
     """
     return cssutils.parseString(css_body, validate=validate)
 
+
+def capitalize_float_margin(css_body):
+    """Capitalize float and margin CSS property names
+    """
+    def _capitalize_property(match):
+        return '{0}:{1}{2}'.format(
+            match.group('property').capitalize(),
+            match.group('value'),
+            match.group('terminator'))
+
+    return _lowercase_margin_float_rule.sub(_capitalize_property, css_body)
+
+
 _element_selector_regex = re.compile(r'(^|\s)\w')
 _cdata_regex = re.compile(r'\<\!\[CDATA\[(.*?)\]\]\>', re.DOTALL)
+_lowercase_margin_float_rule = re.compile(
+    r'''(?P<property>margin(-(top|bottom|left|right))?|float)
+        :
+        (?P<value>.*?)
+        (?P<terminator>$|;)''',
+    re.IGNORECASE | re.VERBOSE)
 _importants = re.compile('\s*!important')
 # These selectors don't apply to all elements. Rather, they specify
 # which elements to apply to.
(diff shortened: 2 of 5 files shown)


JocelynDelalande:
Amended to support (hopefuly) python2.6 and comply with pep8

premailer/premailer.py
```diff
@@ -85,8 +85,28 @@ def _cache_parse_css_string(css_body, validate=True):
...
 
+
+def _capitalize_property(match):
+    return '{0}:{1}{2}'.format(
+        match.group('property').capitalize(),
+        match.group('value'),
+        match.group('terminator'))
+
+
+def _capitalize_float_margin(css_body):
```
peterbe: No need to mark this as private. It's not that secret :)

Also, the fact that you wanted to and were able (I love it!) write a test for this function, also tells me it doesn't need to be a "private" function.

peterbe: Also, don't worry about polluting because define an [`__all__`](https://github.com/oasiswork/premailer/blob/43563fa932a9ccee04218556cbd82c30c6d72c9b/premailer/premailer.py#L38) anyway.

premailer/__main__.py
```diff
@@ -65,6 +65,12 @@ def main(args):
     )
 
     parser.add_argument(
+        "--capitalize-float-margin", default=False,
+        help="Capitalize float and margin properties for outlook.com compat.",
+        action="store_true", dest='capitalize_float_margin'
```
peterbe: quotation mixed use

premailer/premailer.py
```diff
@@ -85,8 +85,28 @@ def _cache_parse_css_string(css_body, validate=True):
     """
     return cssutils.parseString(css_body, validate=validate)
 
+
+def _capitalize_property(match):
```
peterbe: You can move this *into* the closure of `_capitalize_float_margin`.

premailer/premailer.py
```diff
@@ -420,6 +442,14 @@ def transform(self, pretty_print=True, **kwargs):
                 parent = item.getparent()
                 del parent.attrib['class']
 
+        # Capitalize Margin properties
+        # To fix weird outlook bug
+        # https://www.emailonacid.com/blog/article/email-development/outlook.com-does-support-margins
+        if self.capitalize_float_margin:
+            for item in page.xpath('//@style'):
```
peterbe: Is there no better place to do this? What about the code that sets the style attributes in the first place. 
Now, somewhere in premailer it reads the `<style>` and `<link>` and sets the `style` attribute on the DOM and then when it's done that you extract them all out again and re-write them.

premailer/tests/test_premailer.py
```diff
@@ -828,6 +828,43 @@ def test_mailto_url(self):
 
         compare_html(expect_html, result_html)
 
+    def test_uppercase_margin(self):
+        """Option to comply with outlook.
```
peterbe: Can you be more specific and say "outlook.com".

peterbe:
That link, the article, is so interesting that I think you should include it in the test code you wrote. Then people can find out why we have this crazy hack available :)

Great work by the way! I strongly appreciate it.

premailer/premailer.py
JocelynDelalande: Could you be more specific on where to hook-in ? I don't remember where it was exactly in the code but the only places I saw when first looking for somewhere to plug my change were abstracted in a way that didn't let me specify the case of the css property.

Thanks :)

JocelynDelalande:
Thanks for taking the time for feedback :) 

I just updated the PR taking into account all your remarks, except the main one: from where should I apply the modification to CSS property names, waiting for your input on that one, see my other comment.

peterbe:
Re-opening so that Travis runs again. Last time coveralls seems to have gotten stuck.

peterbe:
This will go into the next release (2.10.1).

peterbe:
Included in https://pypi.python.org/pypi/premailer/2.10.1 but not an option on [premailer.io](https://premailer.io) yet.

Issue #143, opened 2015-10-15
Hex color codes are minimised to 3-character if possible


adrianforsius:
6 character color codes are reduced to 3 character color codes if possible e.g:
#FFFFFF => #FFFF
#112233 => #123

This is caused by cssutils parser having this as default behaviour, however this is a common mail-client issue:
![screen shot 2015-10-15 at 09 03 14](https://cloud.githubusercontent.com/assets/616959/10506780/bb16782c-731b-11e5-9ad3-8ceb3a070600.png)

cssutils parser function (cssutils/serialize.py):
```python
def _hash(self, val, type_=None):
        """
        Short form of hash, e.g. #123 instead of #112233
        """
        # TODO: add pref for this!
        if len(val) == 7 and val[1] == val[2] and\
                             val[3] == val[4] and\
                             val[5] == val[6]:
            return u'#%s%s%s' % (val[1], val[3], val[5])
        else:
            return val
```

elidickinson:
Do you have an example of where this causes a color to be misinterpreted? I use premailer and gmail a lot and have never noticed any issues. I think perhaps it was a problem a long time ago, but gmail has long since fixed it. (There is a problem with 3 character background colors and Lotus Notes but we worked around it in #114)

peterbe:
I'd love for there to be a thing in premailer that, by default, tells cssutils to *not* bother.

peterbe:
...because the byte-saving isn't worth it.

adrianforsius:
I made a pull-request to cssutils to have preference for that, in the meanwhile I'll use the fix  discussed in #114 . Thanks for feedback.

peterbe:
@adrianforsius cssutils is no longer maintained. The original author gave me write access and admin access to his PyPI page but I only needed it so I could push a critical fix that makes cssutils work in python 3.5.

peterbe:
Should be solved with https://github.com/peterbe/premailer/commit/e5f73b074a4efb7898204e2fb122b62542494a2e

Issue #114, opened 2015-03-06
Preserve 6 character color codes for Lotus Notes


elidickinson:
As you may be aware, the IBM/Lotus Notes email client has poor HTML support. One of its quirks is that it sometimes (only sometimes!) gets confused by "short" color codes like `#FFF` instead of `#FFFFFF`. Which is unfortunate because Premailer is automatically converting to the short versions when possible. Any idea how to get it to stop? Or even where that's coming from? I'm guessing somewhere deep in cssselect...

This HTML:
```
<table style="background-color:#F00; border-collapse:collapse; position:relative;" bgcolor="#F00" valign="top">
	<tr>
		<td>
			<p style="color:#FFFFFF">This is in a table with background of #F00 (red) </p>
		</td>
	</tr>
</table>


<table style="background-color:#FF0000; border-collapse:collapse; position:relative;" bgcolor="#FF0000" valign="top">
	<tr>
		<td>
			<p style="color:#FFFFFF">This is in a table with background of #FF0000 (red) </p>
		</td>
	</tr>
</table>
```
Looks like this in Notes 8.5:
![screenshot 2015-03-06 16 37 43](https://cloud.githubusercontent.com/assets/44751/6536853/246ae49a-c41f-11e4-948b-c1c71bb0729d.png)

elidickinson:
I *think* Notes is only looking at `bgcolor` attribute anyway, so maybe I can fix this by just tweaking `_style_to_basic_html_attributes` to expand short color codes.

peterbe:
Have you had any luck finding where the conversion of `#ffffff` to `#fff` gets automatically done so we can stop it?

elidickinson:
Well, yes and no. It's somewhere inside `cssutils`:
```
In [10]: cssutils.parseString('h1 { color:#FFFFFF }').cssText
   ....:
Out[10]: 'h1 {\n    color: #FFF\n    }'
```

Though I can hardly blame them since normally I'd prefer the shorter version... it's just problematic with old and broken user agents.

peterbe:
If there's no easy and obvious way to control that from outside `cssutils` then I suspect it's futile and the best course of action would be `Premailer(html).transform().replace('#FFF', '#FFFFFF')` or something smarter.

elidickinson:
For future reference:
```python
# Hacky fix for Lotus Notes. It doesn't handle three character bgcolor codes well
#   see https://github.com/peterbe/premailer/issues/114
[...]

PR #145, opened 2015-10-28
Convert 3-colour codes to 6-colour codes

mpj17:
Dealing with issues like #143 and #114.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -46,6 +46,10 @@ class ExternalNotFoundError(ValueError):
     pass
 
 
+class TransparentIsNotAColour(ValueError):
+    pass
+
+
 def make_important(bulk):
     """makes every property in a string !important.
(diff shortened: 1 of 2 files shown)


mpj17:
Sorry, I never tested in Python 2.6:
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises
"Changed in version 2.7: Added the ability to use assertRaises() as a context manager."

I will change the test to one that does not use a context manager.

peterbe:
2.6 is sad. I wonder if we should make a release that breaks compatibility with 2.6.

mpj17:
*Personally* I have dropped Python 2.6 compatibility. However,  very few people use my code; certainly nothing I write is as widely use as Premailer. For this particular case adding compatibility for 2.6 was easy to do in #147 :)

PR #147, opened 2015-10-28
Dealing with issues like #143 and #114.

mpj17:
Dealing with issues like #143 and #114, this time with flake8 and Python2.6 support!

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -521,14 +521,14 @@ class Premailer(object):
         Notes which fails to handle three character ``bgcolor`` codes well.
         see <https://github.com/peterbe/premailer/issues/114>'''
 
-        shortCodes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
+        short_codes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
         # double digits to enlongen color code
-        retval = shortCodes.sub(r'#\1\1\2\2\3\3', colorValue)
+        retval = short_codes.sub(r'#\1\1\2\2\3\3', colorValue)
 
         # Drop "transparent" bgcolor values entirely
         # The space at the start of the RE is deliberate
-        transparentColor = re.compile(r'transparent', re.I)
-        if transparentColor.match(retval):
+        transparent_color = re.compile(r'transparent', re.I)
+        if transparent_color.match(retval):
             raise TransparentIsNotAColor(retval)
         return retval
 
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -93,7 +93,7 @@ _element_selector_regex = re.compile(r'(^|\s)\w')
 _cdata_regex = re.compile(r'\<\!\[CDATA\[(.*?)\]\]\>', re.DOTALL)
 _importants = re.compile('\s*!important')
 #: The short (3-digit) color codes that cause issues for IBM Notes
-_short_color_codes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
+_short_color_codes = re.compile(r'^#([0-9a-f])([0-9a-f])([0-9a-f])$', re.I)
 
 # These selectors don't apply to all elements. Rather, they specify
 # which elements to apply to.
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -519,10 +519,10 @@ class Premailer(object):
 
     @staticmethod
     def six_color(color_value):
-        '''Fix background colors for Lotus Notes
+        """Fix background colors for Lotus Notes
 
         Notes which fails to handle three character ``bgcolor`` codes well.
-        see <https://github.com/peterbe/premailer/issues/114>'''
+        see <https://github.com/peterbe/premailer/issues/114>"""
 
         # Turn the color code from three to six digits
         retval = _short_color_codes.sub(r'#\1\1\2\2\3\3', color_value)
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -560,11 +560,15 @@ class Premailer(object):
                 attributes['valign'] = value.strip()
             elif key == 'background-color':
                 try:
-                    sixcolor = self.six_color(value.strip())
+                    six_color = self.six_color(value.strip())
                 except TransparentIsNotAColor:
+                    # The color "transparent" is interpreted by IBM Notes as
+                    # black, so just ignore the color and hope. The right thing
+                    # to do is render the element in an actual browser and
+                    # figure out what color is actually beneath this element.
                     pass
                 else:
-                    attributes['bgcolor'] = sixcolor
+                    attributes['bgcolor'] = six_color
             elif key == 'width' or key == 'height':
                 value = value.strip()
                 if value.endswith('px'):
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -518,14 +518,14 @@ class Premailer(object):
         return css_body
 
     @staticmethod
-    def six_color(colorValue):
+    def six_color(color_value):
         '''Fix background colors for Lotus Notes
 
         Notes which fails to handle three character ``bgcolor`` codes well.
         see <https://github.com/peterbe/premailer/issues/114>'''
 
-        # double digits to enlongen color code
-        retval = _short_color_codes.sub(r'#\1\1\2\2\3\3', colorValue)
+        # Turn the color code from three to six digits
+        retval = _short_color_codes.sub(r'#\1\1\2\2\3\3', color_value)
 
         # Drop "transparent" bgcolor values entirely
         # The space at the start of the RE is deliberate
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -548,9 +548,9 @@ class Premailer(object):
                 attributes['align'] = value.strip()
             elif key == 'vertical-align':
                 attributes['valign'] = value.strip()
-            elif ((key == 'background-color') 
+            elif ((key == 'background-color')
                   and ('transparent' not in value.lower())):
-                # Only add the 'bgcolor' attribute if the value does not 
+                # Only add the 'bgcolor' attribute if the value does not
                 # contain the word "transparent"; before we add it possibly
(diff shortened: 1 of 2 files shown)


mpj17:
CI passes!

In summary, the `six_colour` static method that expands a three-digit colour value to six. If given the colour `transparent` is given it raises an exception. (It is a static method because it makes the unit testing much easier.)

Both the return value from `six_colour` method and the `TransparentIsNotAColour` exception is handled by the existing `_style_to_basic_html_attributes` method. In the case of the exception nothing is set. I *think* this is right, but I lack IBM Notes to do a good test.

peterbe:
Would you mind rebasing against master? 
As of https://github.com/peterbe/premailer/pull/148 the coverage testing should no longer hang here in GitHub.

peterbe:
Also, would you mind calling it "color" instead of "colour"? The word "color" is used throughout the code and the reason it's spelled like that is because CSS is based on American English. Consistency is important.

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
...
+        shortCodes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
+        # double digits to enlongen color code
+        retval = shortCodes.sub(r'#\1\1\2\2\3\3', colourValue)
+
+        # Drop "transparent" bgcolor values entirely
+        # The space at the start of the RE is deliberate
+        transparentColor = re.compile(r'transparent', re.I)
+        if transparentColor.match(retval):
+            raise TransparentIsNotAColour(retval)
+        return retval
```
peterbe: Would you mind pep8 fix this? E.g instead of "shortCodes", make it "short_codes"

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
...
         return css_body
 
+    @staticmethod
+    def six_colour(colourValue):
+        '''Fix background colours for Lotus Notes
+
+        Notes which fails to handle three character ``bgcolor`` codes well.
+        see <https://github.com/peterbe/premailer/issues/114>'''
+
+        shortCodes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
```
peterbe: Move this to global somewhere [around line 90-100](https://github.com/mpj17/premailer/blob/master/premailer/premailer.py#L91-L98). Otherwise it has to be compiled for every call of this static method.

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
...
+        # double digits to enlongen color code
```
peterbe: "enlongen"??

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
...
+        retval = shortCodes.sub(r'#\1\1\2\2\3\3', colourValue)
+
+        # Drop "transparent" bgcolor values entirely
+        # The space at the start of the RE is deliberate
```
peterbe: What space?

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
 
         return css_body
 
+    @staticmethod
+    def six_colour(colourValue):
+        '''Fix background colours for Lotus Notes
+
+        Notes which fails to handle three character ``bgcolor`` codes well.
+        see <https://github.com/peterbe/premailer/issues/114>'''
```
peterbe: The convention is to use `"""` not `'''`

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
...
+
+        shortCodes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
+        # double digits to enlongen color code
+        retval = shortCodes.sub(r'#\1\1\2\2\3\3', colourValue)
+
+        # Drop "transparent" bgcolor values entirely
+        # The space at the start of the RE is deliberate
+        transparentColor = re.compile(r'transparent', re.I)
+        if transparentColor.match(retval):
```
peterbe: No need to compile a regular expression for that. This would be enough:
```
if 'transparent' in reval.lower():
```

premailer/premailer.py
peterbe: If you use `re.I` you don't need to type `A-F`. The default thing is to write in lowercase.

premailer/premailer.py
```diff
@@ -510,6 +514,24 @@ def _load_external(self, url):
...
+        '''Fix background colours for Lotus Notes
+
+        Notes which fails to handle three character ``bgcolor`` codes well.
+        see <https://github.com/peterbe/premailer/issues/114>'''
+
+        shortCodes = re.compile(r'^#([0-9A-F])([0-9A-F])([0-9A-F])$', re.I)
+        # double digits to enlongen color code
+        retval = shortCodes.sub(r'#\1\1\2\2\3\3', colourValue)
+
+        # Drop "transparent" bgcolor values entirely
```
peterbe: Where does it "drop"?

premailer/premailer.py
```diff
@@ -535,7 +557,12 @@ def _style_to_basic_html_attributes(self, element, style_content,
             elif key == 'vertical-align':
                 attributes['valign'] = value.strip()
             elif key == 'background-color':
-                attributes['bgcolor'] = value.strip()
+                try:
+                    sixColour = self.six_colour(value.strip())
+                except TransparentIsNotAColour:
+                    pass
```
peterbe: What is that allowed to pass?

peterbe: oh. I think I see. It's because you have an `else:` next. 
Here's what I suggest you do instead. 

1) Remove the `transparent` check from the `six_colour` function. That way, that function does 1 thing and 1 thing only. 
2) Remove the `TransparentIsNotAColour` exception class
3) Simply do this instead of using exceptions:
```python
if 'transparent' not in value.lower():
    # Only add the 'bgcolor' attribute if the value doesn't contain the word "transparent"
    # and before we add it possibly correct 3 character shortenings to 6 character
    # equivalents. E.g. convert "abc" to "aabbcc".
    attributes['bgcolor'] = self.six_colour(value.strip())
```

premailer/tests/test_premailer.py
```diff
@@ -2501,3 +2502,41 @@ def test_remove_unset_properties(self):
...
+        expect_html = """<html>
+  <head>
+  </head>
+  <body style="background-color:#fe5" bgcolor="#ffee55">
+    <h1 style="color:#f0df0d">Colour test</h1>
+    <p style="background-color:#123456" bgcolor="#123456">
+      This is a test of colour handling.
+    </p>
+  </body>
+</html>"""
```
peterbe: Most of the tests in this file try to stick to a convention of indenting the multi-line HTML strings by 4 spaces.

premailer/tests/test_premailer.py
```diff
@@ -2501,3 +2502,41 @@ def test_remove_unset_properties(self):
...
+    p {background-color: #123456;}
+    h1 {color: #f0df0d;}
+  </style>
+  <body>
+    <h1>Colour test</h1>
+    <p>
+      This is a test of colour handling.
+    </p>
+  </body>
+</html>'''
```
peterbe: Use `"""` instead.

mpj17:
Done.

premailer/premailer.py
```diff
@@ -510,6 +513,17 @@ def _load_external(self, url):
...
+    @staticmethod
+    def six_color(color_value):
+        """Fix background colors for Lotus Notes
+
+        Notes which fails to handle three character ``bgcolor`` codes well.
+        see <https://github.com/peterbe/premailer/issues/114>"""
+
+        # Turn the color code from three to six digits
+        retval = _short_color_codes.sub(r'#\1\1\2\2\3\3', color_value)
+        return retval
```
peterbe: No need to create the variable if you're immediately returning it.

premailer/premailer.py
```diff
@@ -534,8 +548,13 @@ def _style_to_basic_html_attributes(self, element, style_content,
                 attributes['align'] = value.strip()
             elif key == 'vertical-align':
                 attributes['valign'] = value.strip()
-            elif key == 'background-color':
-                attributes['bgcolor'] = value.strip()
+            elif ((key == 'background-color')
+                  and ('transparent' not in value.lower())):
```
peterbe: No need for these extra brackets.
```python
elif key == 'background-color' and 'transparent' not in value.lower():
```
or, if that's too long:
```python
elif (
    key == 'background-color' and 
    'transparent' not in value.lower()
):
```

premailer/premailer.py
```diff
@@ -534,8 +548,13 @@ def _style_to_basic_html_attributes(self, element, style_content,
...
+                # Only add the 'bgcolor' attribute if the value does not
+                # contain the word "transparent"; before we add it possibly
+                # correct the 3-digit color code to its 6-digit equivalent
+                # ("abc" to "aabbcc") so IBM Notes copes.
```
peterbe: Excellent!

peterbe:
Great work! Just a few small last tweaks. Would you mind rebasing it into 1 commit after?

premailer/premailer.py
mpj17: Feel free to change it. I prefer setting `retval` as it makes debugging much easier later on. TBH there really should be an `assert` in there:

    assert len(retval) == 7, 'Color value ({0}) of the wrong length: {0}'.format(retval, len(retval))

premailer/premailer.py
mpj17: I have never found a Boolean expression without brackets that I have ever trusted. Feel free to change it.

peterbe:
Hey @mpj17 If you don't mind, I'm going to rebase and tidy up the last little niggles on this PR.

peterbe:
Having said that, I need to wait for https://github.com/peterbe/premailer/pull/154 because at the moment our test suite pulls down a version of lxml for pypy that pypy can't build. So all PRs are failing.

mpj17:
No worries what so ever :)

peterbe:
Branch rebased and moved to https://github.com/peterbe/premailer/pull/155

PR #158, opened 2016-01-19
Add Python 3.5 to tox and travis and drop Python 3.2 support

pzrq:
Works out of the box for me on OSX El Capitan, let's see what Travis says.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ env:
   - TOX_ENV=flake8
   - TOX_ENV=py26
   - TOX_ENV=py27
-  - TOX_ENV=py32
   - TOX_ENV=py33
   - TOX_ENV=py34
   - TOX_ENV=py35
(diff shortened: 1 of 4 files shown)


peterbe:
Would you mind squashing that bug into 1 commit?

pzrq:
Hmm, thoughts on what you'd like to do? 
https://github.com/pypa/pip/commit/b11cb019a47ff0cf3d8a37a0c89d8ae4cf25282f

peterbe:
Then I think we should drop travis support for 3.2 too. That release is getting really old. 

Please add another commit to this PR that drops 3.2 from `tox.ini` and `.travis.yml` and ping me for a merge.

pzrq:
@peterbe Should be good to go with dropping Python 3.2 support.

peterbe:
Ahhh nice!

pzrq:
Thanks @peterbe

PR #167, opened 2016-06-07
keep classes by default, fixes #33

### LINKED ISSUE #33: keep class attributes
the transform() function replaces the "class" attribute with the appropriate styling, but it also removes the class from the html elements. I need the styling added inline, but the classes to be left within their elements so that my @media (!important) styling can be inserted in after and still have classes to point to within the dom.

Im having a hard time changing your code to meet this needs, do you have any thoughts on how I could keep the classes within the html elements but still inject the inline styling?

peterbe:
Major version bump.

### DIFF
--- a/premailer/__init__.py
+++ b/premailer/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import, unicode_literals
 from .premailer import Premailer, transform
 
-__version__ = '2.11.0'
+__version__ = '3.0.0'
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -121,7 +121,7 @@ class Premailer(object):
                  exclude_pseudoclasses=True,
                  keep_style_tags=False,
                  include_star_selectors=False,
-                 remove_classes=True,
+                 remove_classes=False,
                  capitalize_float_margin=False,
                  strip_important=True,
                  external_styles=None,
--- a/premailer/tests/test_premailer.py
+++ b/premailer/tests/test_premailer.py
@@ -1088,7 +1088,7 @@ b
         <head>
         </head>
         <body>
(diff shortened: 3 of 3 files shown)


graingert:
I think it's such a trivial and unlikely to be used piece of code, that I'd rather just remove it and recommend people add their own:
```python
            # now we can delete all 'class' attributes
            for item in page.xpath('//@class'):
                parent = item.getparent()
                del parent.attrib['class']
```

peterbe:
@graingert But I want the library to be as easy as possible. Also, as lazy as possible :)

graingert:
It's your call, but imagine if this feature was not implemented, would it be worth adding?

I don't see a use-case for it.

peterbe:
One usecase is that it's not hard to keep and it's nice to have it there
for people who needed it.

Actually, if you don't have media queries. If you have some really simple
CSS for your HTML emails, then having the `class` attributes still in means
the email gets unnecessarily larger.
I suggest we just keep the functionality around.

On Tue, Jun 7, 2016 at 8:45 AM, Thomas Grainger <notifications@github.com>
wrote:

> It's your call, but imagine if this feature was not implemented, would it
> be worth adding?
>
> I don't see a use-case for it.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/peterbe/premailer/pull/167#issuecomment-224269061>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AABoc_2Fg2_REhvGVNqOZLI7Z5RBUWkLks5qJWfjgaJpZM4Iv3_H>
> .
>



-- 
Peter Bengtsson
Mozilla Tools & Services

Issue #93, opened 2014-11-21
Output replaces HTML Entities with unicode literals


elidickinson:
Running transform seems to translate HTML entities in the source into unicode literals. For example:
`<p>&copy; &nbsp;&nbsp; 2014</p>`
becomes
`<p>©    2014</p>`

This is causing issues for me and I'm guessing it's just a side effect of the lxml settings and not intentional. My understanding is that "&copy" has better email client compatibility as "©" (If anything I'd prefer an option to go the other way: escape any unicode literals in the source)

dandv:
Running `Copyright &copy; 2015 iDoRecall, Inc.` through premailer.io, I saw `&copy;` converted to just "c", not the Unicode copyright symbol:

> Copyright c 2015 iDoRecall, Inc.

Definitely need an option against that transformation.

peterbe:
If there's a way to tell lxml to stay the heck out of trying to do that trick that would be great. And something we ought to do by default.

1951FDG:
Noticed that ```&rsquo;``` get converted as well as ```&quot;```

OrangeDog:
> a way to tell lxml

Maybe `resolve_entities=False`?

Issue #157, opened 2016-01-18
Mojibake: With unicode crescent moon 🌙 0x1f319


pzrq:
Might be similar to #142, #152.

```
$ python
Python 3.4.3 (default, Jan  5 2016, 10:40:34) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from premailer.premailer import Premailer
>>> html = '''
... <html xmlns="http://www.w3.org/1999/xhtml">
... <body>
...             Dear Skyelar 🌙,
... </body>
... </html>
... '''
>>> p = Premailer(html)
>>> mojibake = p.transform()
>>> print(mojibake)
<html>
<head></head>
<body><p>h   t   m   l       x   m   l   n   s   =   "   h   t   t   p   :   /   /   w   w   w   .   w   3   .   o   r   g   /   1   9   9   9   /   x   h   t   m   l   "   &gt;   
   </p></body>
</html>

>>> 
```

pzrq:
Have put this into a test into the following branch: https://github.com/peterbe/premailer/compare/master...mathspace:unicode-crescent-moon

For me, it works fine under Python 2.6 and 2.7 tox environments, but fails under Python 3.4 and 3.5 tox environments. 

It fails at `etree.tostring(root, **kwargs)` suggesting this issue at least is an `lxml` bug.

peterbe:
What's so special about that emoji specifically? Is it not a problem with other emojis?

pzrq:
@peterbe It was the emoji that came up in our issue tracker in a staging environment. Nothing particularly special and the issue may well affect a significant set of unicode characters, e.g. (untested) perhaps those needing more than 16 bits to encode in utf-8.

peterbe:
@pzrq Bummer. Why don't you put up a PR with your work so far and perhaps other people can comment to help make it work in all environments.

pzrq:
Closing due to inactivity.

Issue #152, opened 2016-01-08
UnicodeEncodeError: 'ascii' codec can't encode character


oldo:
Hi Peter,
I've created an html document with bootstrap styling and when I try and run Premailer from the CLI I get the following error:

```
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/oldo/Python/virtual-environments/AMS-Journal/lib/python2.7/site-packages/premailer/__main__.py", line 142, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Users/oldo/Python/virtual-environments/AMS-Journal/lib/python2.7/site-packages/premailer/__main__.py", line 137, in main
    options.outfile.write(p.transform(pretty_print=options.pretty))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2013' in position 106688: ordinal not in range(128)
```
I have tried everything that I can think of to try and avoid this, but can't seem to make any progress. What would usually cause this error and do you know a way to fix it?
Thanks!

oldo:
I was finally able to find a workaround to avoid this error :-)
Please [see here](http://stackoverflow.com/questions/34681040/premailer-unicodeencodeerror-ascii-codec-cant-encode-character-u-u2013-i/34691698#34691698) if you are interested.

peterbe:
How do you even reproduce this error?
premailer has a pretty extensive set of unit tests and they cover much unicode handling stuff. 

A lot of times, the reason you get into UnicodeEncodeError is because of something going wrong *before* you use the library or third-party function that is raising the error. For example, in python 2:
```python
>>> def foo(add):
...     return u'a' + add
...
>>>
```
Seems sensible, right? It adds the (unicode) character `a` to the start of whatever you put in:
```python
inp = 'ä'
>>> foo(inp)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in foo
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
```
What happened here is that it was a mistake to define the variable `inp` as a byte string when it contained non-ascii characters. 

[...]

PR #169, opened 2016-06-15
Allow setting encoding from command-line.

OrangeDog:
Passing `--encoding ascii` should fix issues #93, #100, #152 and #157.

### DIFF
--- a/premailer/__main__.py
+++ b/premailer/__main__.py
@@ -112,6 +112,11 @@ def main(args):
         action="store_true",
         help="Pretty-print the outputted HTML.",
     )
+    
+    parser.add_argument(
+        "--encoding", default='utf-8',
+        help="Output encoding. The default is utf-8",
+    )
 
     options = parser.parse_args(args)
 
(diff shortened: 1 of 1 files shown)


OrangeDog:
I'm not sure exactly what the CI is complaining about.
If someone with more experience than I could fix it then that would be great.

premailer/__main__.py
```diff
@@ -140,7 +145,7 @@ def main(args):
         disable_basic_attributes=options.disable_basic_attributes,
         disable_validation=options.disable_validation
     )
-    options.outfile.write(p.transform(pretty_print=options.pretty))
+    options.outfile.write(p.transform(encoding=options.encoding, pretty_print=options.pretty))
```
peterbe: CI fails because this line is not PEP8 compliant. You need to re-write it so it's not longer than 80 characters. 
Change it to be indented like this:
```python
options.outfile.write(p.transform(
    encoding=options.encoding, 
    pretty_print=options.pretty
))
```

premailer/__main__.py
```diff
@@ -112,6 +112,11 @@ def main(args):
         action="store_true",
         help="Pretty-print the outputted HTML.",
     )
+
```
peterbe: This line has excessive whitespace. 
Blank lines should only be a single newline character.

peterbe:
I'll prepare a new release based on this.

peterbe:
[3.0.1](https://pypi.python.org/pypi/premailer/3.0.1)

Issue #130, opened 2015-07-17
Mailchimp merge vars are broken


dandv:
MailChimp merge vats have the syntax `*|VARIABLENAME|*`. The problem is that the `|` character is replaced by premailer with %7C, so MailChimp won't parse the variables.

peterbe:
Where do you put those variables? Inside tags (as attribute values) or inside block tags (as stuff between `>` and `<`)?

elidickinson:
They can technically appear anywhere. It's kind of a tricky one, because premailer isn't *wrong* for trying to escape non-standard characters and other ESPs use other weird characters: I think Campaign Monitor is `%VARIABLE%` and Sailthru is `{VARIABLE}`

peterbe:
@dandv The immediate fix is to do something like:

```
p=Premailer(html)
mailchimp.send(p.transform().replace('%2A%7CVARIABLENAME%7C%2A', '*|VARIABLENAME|*))
```
I guess you might need to rewrite that to a regular expression or something.

peterbe:
By the way I tried pasting in this html
```
<html>
<style>
p { color:red }
</style>
<body>
<p>*|PARAGRAPH|*</p>
</body></html>
```
into [Premailer.io](http://premailer.io) and it came out like this:
```
<html>
<head></head>
<body>
<p style="color:red">*|PARAGRAPH|*</p>
</body>
</html>
```

minichate:
An example where the merge variable gets escaped is:

```html
<html>
<body>
<img src="*|MERGEVAR|*" />
</body>
</html>
```

I'd love to submit a pull request to fix this behaviour! Any hints where I might start looking?

peterbe:
@minichate I suspect this gets escaped by lxml. And if so, I think there's a good reason for it. 
I think you're better off doing a string replacement like mentioned in my comment above.

lavr:
Guys, if you're searching where to start, here is a point:

    >>> from premailer import Premailer
    >>> from lxml import etree
    >>> etree.tostring(Premailer(html=etree.fromstring('<img src="*|MERGEVAR|*" />', parser=etree.HTMLParser())).transform(), method='xml')
    '<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">\n<html><head/><body><img src="*|MERGEVAR|*"/></body></html>'

[...]

Issue #72, opened 2014-07-08
Jinja variables in hrefs are HTML encoded. Outside of hrefs they're not encoded.


skyler:
Premailer version 2.5.0.

I have an HTML template with [Jinja](http://jinja.pocoo.org/) variables in it. I'd like for these Jinja variables to remain after using premailer, so I can cache the premailed version and do variable substitution in a loop.

For variables in hrefs, like
 `<a href="{{ SURVEY_LINK }}" class="btn-primary">Start Survey</a>`

Premailer HTML encodes the braces to their HTML-encoded entities:
 `<a href="%7B%7B%20SURVEY_LINK%20%7D%7D" class="btn-primary"...`

Likewise, brackets, `[` and `]` in href quotes are escaped, which is problematic for my use case as well.

I'd like to use premailer on my Jinja email templates but because of this HTML encoding issue I'm not able to.

clj:
I have not had much luck using the xml output method. While it avoids escaping the attributes, I have just get various XML related errors instead, primary related non declared entities. Instead, I decided to write my own version of etree's `tostring` which renders the HTML and keeps the values in the attributes unescaped.

The following gist has this code: https://gist.github.com/clj/b8ba315b9a138db73be0

This works well for my testcase, but I have not tested it extensively as I am having some other problems (where the generated output as rendered by browsers is not equivalent to the input, but I don't think this is due to the renderer linked above, but I am investigating). Basically, YMMV.

Getting the output rendered as I want requires overriding and copying some of the functionality of `transform` in the `premailer` class and it might be nice to refactor the code so that is possible to easily override the rendering of the output without doing that.

peterbe:
Perhaps the best way is to allow you to insert some sort of override for how lxml puts the values into `src` and `href` attributes. Premailer fiddles with the value every time because it tries to correct URLs based on the `base_url`. 
See https://github.com/peterbe/premailer/blob/d2a2a4afdd06e5931545bc76b9ee940fdefdb543/premailer/premailer.py#L421

In your case, you don't want an actual working browser URL in there, you want some other code. 

Perhaps something like this:
```python
def my_url_fixed(old, new):
    if '{{'  in old and '}}' in old:
        # I know what I'm doing :)
        return old
    return new

p=Premailer(jinja_html, base_url='https://example.com', url_fixer=my_url_fixer)
```
And inside Premailer we have something like this:
```python
    if self.url_fixer is None:
      self.url_fixer = lambda _, n: n
[...]

PR #171, opened 2016-06-15
Add CLI option to exclude strings from encoding

OrangeDog:
Intended as an illustration of a solution to issues #72 and #130 (and also #93 et al.)
I assume this actually wants pushing into `transform` somewhere.


firstcloudconsulting:
+1

PR #173, opened 2016-08-09
Encode transformed before writing to file

bonya:
Python2 cli support for no-ascii characters when writing to output file


peterbe:
This needs a bit more work. 
It needs to work for python2 and python3 and be fully tested.

peterbe:
Too old. Sorry.

PR #176, opened 2016-09-20
Raising an error on external style 404s

rclmenezes:
If you give premailer an external stylesheet that 404s, it should raise an ExternalNotFoundError. Otherwise, it will attempt to parse the 404 page as a CSS file and give you some funky errors.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -505,7 +505,7 @@ class Premailer(object):
 
     def _load_external_url(self, url):
         r = requests.get(url)
-        if r.status_code == 400:
+        if r.status_code == 404:
             raise ExternalNotFoundError(url)
         return r.text
 
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -504,7 +504,10 @@ class Premailer(object):
             return out
 
     def _load_external_url(self, url):
-        return requests.get(url).text
+        r = requests.get(url)
+        if r.status_code == 400:
+            raise ExternalNotFoundError(url)
+        return r.text
 
     def _load_external(self, url):
         """loads an external stylesheet from a remote url or local path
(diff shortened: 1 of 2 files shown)


premailer/premailer.py
```diff
@@ -504,7 +504,10 @@ def transform(self, pretty_print=True, **kwargs):
             return out
 
     def _load_external_url(self, url):
-        return requests.get(url).text
+        r = requests.get(url)
+        if r.status_code == 400:
```
peterbe: Shouldn't that be 404?

premailer/tests/test_premailer.py
```diff
@@ -1828,6 +1829,19 @@ def test_load_external_url(self, mocked_requests):
         mocked_requests.get.assert_called_once_with(faux_uri)
         eq_(faux_response, r)
 
+    @mock.path('premailer.premailer.requests')
```
peterbe: Should that be `@mock.patch(...)`?

rclmenezes: Ha! I got too lazy to run the tests. tox was giving me trouble. I'll fix them and run the tests tomorrow :)

peterbe:
This might need to be optional. I.e. make it so that instead of trying to parse the response body during a 404 it **ignores** it. Perhaps with a warning. That's what `cssutils` does when it encounters a bad CSS selector it doesn't understand. 
However, raising an error is probably a good default. 

Also, I wonder if the conditional should instead be `if r.status != 200: raise ...`. 
A CSS file might sit on a CDN or something that is temporarily down. What do you think?

rclmenezes:
Hmmm...

I would avoid `status != 200` because I could imagine a 301 or 302 being a valid use case.

Personally, I'd rather have an error in these types of situations. I can't imagine a situation where someone *wants* a 404. Alternatively, we could check the Content-Type and send a warning when it's not text/css? After all, most browsers won't interpret them as CSS stylesheets anyway with the wrong mime-type.

peterbe:
Perhaps instead of `!200` we could do `if status >= 400: raise ...`
What do you think of that?

rclmenezes:
That sounds reasonable!

peterbe:
Closing in favor of https://github.com/peterbe/premailer/pull/212

PR #183, opened 2017-07-03
Do not append base_url to tel: urls

poxip:
"tel:" links should be left as they are, withouth prepending base_url (this fixes: #140)

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -485,6 +485,8 @@ class Premailer(object):
                         url.startswith('cid:')
                     ):
                         continue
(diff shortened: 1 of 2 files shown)


peterbe:
I guess you want a new release that contains this. Coming up.

peterbe:
See https://github.com/peterbe/premailer/commit/597b57740c661df8b3f4f5bdec7c495afe955275 and https://pypi.python.org/pypi/premailer/3.1.0

poxip:
Great, thank you!

PR #188, opened 2018-04-03
Add option to preserve all links in html

mokibit:
We are trying to use this great module in cppreference-doc project for creating browsable HTML documentation for various help viewers. One of them does not support external CSS nor style tags, thus your project is very handy.

Unfortunately current URL processing in premailer breaks relative links between pages. I've added an option to disable this. Can this change be included into the official premailer project?

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -116,6 +116,7 @@ class Premailer(object):
     attribute_name = 'data-premailer'
 
     def __init__(self, html, base_url=None,
+                 preserve_all_links=False,
                  preserve_internal_links=False,
                  preserve_inline_attachments=True,
                  exclude_pseudoclasses=True,
@@ -138,6 +139,14 @@ class Premailer(object):
                  remove_unset_properties=True):
         self.html = html
         self.base_url = base_url
+
+        # if base_url is specified, premailer will transform all URLs by joining
+        # them with the base_url. Setting preserve_internal_links to True
+        # will disable this behavior for links to named anchors. Setting
+        # preserve_inline_attachments to True will disable this behavior for
+        # any links with cid: scheme. Setting preserve_all_links to True will
+        # disable this behavior altogether.
+        self.preserve_all_links = preserve_all_links
         self.preserve_internal_links = preserve_internal_links
         self.preserve_inline_attachments = preserve_inline_attachments
         self.exclude_pseudoclasses = exclude_pseudoclasses
@@ -473,7 +482,7 @@ class Premailer(object):
         #
         # URLs
         #
-        if self.base_url:
+        if self.base_url and not self.preserve_all_links:
             if not urlparse(self.base_url).scheme:
                 raise ValueError('Base URL must have a scheme')
             for attr in ('href', 'src'):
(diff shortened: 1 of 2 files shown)


mokibit:
All tests suites pass except one which fails due to reasons unrelated to my PR (missing pypy interpreter).

p12tic:
Hey @peterbe, could you please look into this PR?

We would like to use premailer in the [cppreference-doc](https://github.com/p12tic/cppreference-doc) project and the functionality added in this PR is the only thing missing. We need this change in official premailer sources because cppreference-doc is a Debian package, so in order to use premailer we will need to package it for Debian too and extra non-bugfix patches will make the process much harder.

I believe that the pypy issue in tests has been fixed in https://github.com/peterbe/premailer/pull/190.

premailer/premailer.py
```diff
@@ -138,6 +139,14 @@ def __init__(self, html, base_url=None,
                  remove_unset_properties=True):
         self.html = html
         self.base_url = base_url
+
+        # if base_url is specified, premailer will transform all URLs by
+        # joining them with the base_url. Setting preserve_internal_links to
+        # True will disable this behavior for links to named anchors. Setting
+        # preserve_inline_attachments to True will disable this behavior for
+        # any links with cid: scheme. Setting preserve_all_links to True will
+        # disable this behavior altogether.
```
peterbe: This is a good comment but my mind immediately wonders; Why not just *not* set `base_url`?

premailer/premailer.py
```diff
@@ -116,6 +116,7 @@ class Premailer(object):
     attribute_name = 'data-premailer'
 
     def __init__(self, html, base_url=None,
+                 preserve_all_links=False,
```
peterbe: I don't like the name. Because the intention of this argument is to **disable** the functionality of messing with links. Perhaps something like `disable_link_rewrites=False`. Does that read better to you?

premailer/premailer.py
peterbe: I had to read the code to answer that. The answer is that if you set `base_url` it is used for two things:

1. Rewrite relative links.
2. Correct relative URLs for loading external style sheets. 

The comment should mention that the `base_url` is used for fetching external style sheets as well.

premailer/premailer.py
peterbe: Or `disable_link_rewrites=False`?

Whatever we use, I like booleans to default to False as the good default.

peterbe:
I totally understand your use case. It's a great idea. As soon we manage to figure this out, I'll make a new release and you can start using it in cppreference-doc.

premailer/premailer.py
mokibit: Yes, that indeed reads better. I will fix that.

premailer/premailer.py
mokibit: Agreed.

mokibit:
I think I addressed your comments. Does it look good? Do you want me to merge the commits?

peterbe:
https://github.com/peterbe/premailer/commit/ebfd310dbd0a88b465e811411f67a360e11f3292

Now, if I can just remember how to make a PyPI release for this project.

PR #197, opened 2018-09-27
Added support for `data-premailer="ignore"` HTML attribute

### LINKED ISSUE #196: Ignore specific link tag or style block?
Is it possible to ignore a specific link tag or style block when running HTML through Premailer?

My use case is that I want to include a custom `@font-face` from Google, but of course the `@font-face` logic gets stripped by Premailer (since there's nothing to attach it to when inlining). I went through the code and couldn't find any way to get Premailer to ignore this `<link>` tag aside from leaving off the `rel` attribute (which would presumably lead to rendering issues depending on browser/client).

onecrayon:
`<style>` and `<link>` tags with this attribute will be ignored by Premailer, allowing you to skip inlining for things like Google fonts stylesheets. This is the method used by the Ruby Premailer script:

https://github.com/premailer/premailer/#premailer-specific-css

Closes #196.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -329,7 +329,11 @@ class Premailer(object):
         rules = []
         index = 0
 
-        for element in CSSSelector('style,link[rel~=stylesheet]')(page):
+        selectors = [
+            'style:not([data-premailer="ignore"])',
+            'link[rel~=stylesheet]:not([data-premailer="ignore"])'
+        ]
+        for element in CSSSelector(','.join(selectors))(page):
             # If we have a media attribute whose value is anything other than
             # 'all' or 'screen', ignore the ruleset.
             media = element.attrib.get('media')


peterbe:
Great start! I don't think this will be merged until the README is updated and a passing unit test. Looking forward to see it.

onecrayon:
I went to add a test and update the Readme and was surprised to find pre-existing tests, and the ReadMe already claiming this was possible. I'm not sure why it wasn't working when I initially tested it, or how I missed the documentation in the ReadMe, but this is evidently already supported (and worked perfectly when I tested it today).

Sorry for wasting your time!

peterbe:
> Sorry for wasting your time!

Not at all! I'm sorry the documentation and the API wasn't clear that this already was possible. 
Do you think it was "too hidden" for the next person, like you, you have this question. Perhaps something can be made loud and bold in the README.

onecrayon:
I think what happened was I hit "For more advanced options, check out the code of the Premailer class and all its options in its constructor" and the section on command-line usage, and failed to progress further down the readme (instead went haring off into the codebase). If I'd scrolled down a bit more the section on excluding tags was pretty well highlighted. About the only way it would be more obvious would be if the initial code example included an ignored section, but that might be overkill.

peterbe:
One idea is to document it more clearly like this:

```python
>>> from premailer import transform
>>> print transform("""
        <html>
        <style type="text/css">
        h1 { border:1px solid black }
        p { color:red;}
        p::first-letter { float:left; }
        </style>
        <h1 style="font-weight:bolder">Peter</h1>
        <p>Hej</p>
        </html>
        """,
        # Demonstrating all the possible options and their defaults...
        # base_url=None,
        # disable_link_rewrites=False,
        # preserve_internal_links=False,
        ...
        
)
```
If you could take a stab at doing that, in the README, in a PR I'd love to take a look. 

Also, sorry that the options' defaults are such a mess. Old (mature?) project with many contributors over the years.

peterbe:
Also, if you make a PR like that, can you change to `print()` instead of the Python2 style with `print` call.

PR #198, opened 2018-10-04
Add caching to more functions

p12tic:
On [cppreference-doc](https://github.com/p12tic/cppreference-doc) we are processing around 4000 comparatively large pages using premailer. This takes a while with the current implementation. After looking into profiler results and doing some experimental changes it turns out that simply caching results of more functions and increasing the size of the caches speed up premailer by around 2.5 times.

This PR implements the above-mentioned changes in a way that the new behavior can be enabled on runtime explicitly, old-behavior is default and existing users are almost not impacted.

The `function_cache` decorator is modified to accept the size of the cache as a argument to the wrapper, so that it can be changed on runtime. The size of the caches may now be controlled via additional parameter to the `Premailer.__init__` method.

Also, `function_cache` is modified not to drop the contents of the cache when it overflows: this is the canonical behavior of caches. Small cache may still be very useful for handling the majority of requests and since it already reached the maximum size without inducing out of memory condition, being at the maximum size instead of zero will likely not induce out of memory condition later. If users are concerned about an out of memory condition arising later, when doing other tasks, we should provide a method to clear caches explicitly.

Furthermore, `function_cache` is modified to convert `list` arguments to `tuples` so that they can be hashed. `_HashedSeq` can be retired this way.

Caching behavior has been added to the top functions that I saw in the profiling results. The cache itself has very low overhead, so we know that at worst case we will only increase memory usage without reducing performance too much.

Finally, python 2.6 and 3.3 have been removed from the travis.yml as the upstream testing tools no longer support these python versions.

Please let me know what do you think about this. Thanks for your time!

### DIFF
--- a/premailer/cache.py
+++ b/premailer/cache.py
@@ -21,31 +21,21 @@ class _Cache(object):
         self.cache = {}
 
 
-def function_cache(expected_max_entries=1000):
+def function_cache():
     """
-        function_cache is a decorator for caching function call
-        the argument to the wrapped function must be hashable else
+        function_cache is a decorator for caching function call.
+        The argument to the wrapped function must be hashable else
         it will not work
 
-        expected_max_entries is for protecting cache failure. If cache
-        misses more than this number the cache will turn off itself.
-        Specify None you sure that the cache  will not cause memory
-        limit problem.
-
-        Args:
-            expected_max_entries(integer OR None): will raise if not correct
+        Function wrapper accepts max_cache_entries argument which is for
+        protecting against cache failure. If cache misses more than this number
+        the cache will turn off itself. Specifying None will not limit cache
+        at all.
 
         Returns:
             function
 
     """
-    if (
-        expected_max_entries is not None and
-        not isinstance(expected_max_entries, int)
-    ):
-        raise TypeError(
-            'Expected expected_max_entries to be an integer or None'
-        )
 
     # indicator of cache missed
     sentinel = object()
@@ -55,7 +45,17 @@ def function_cache(expected_max_entries=1000):
 
         @functools.wraps(func)
         def inner(*args, **kwargs):
-            if cached.off:
+            max_cache_entries = kwargs.pop('max_cache_entries', 1000)
+
+            if (
+                max_cache_entries is not None and
+                not isinstance(max_cache_entries, int)
+            ):
+                raise TypeError(
+                    'Expected expected_max_entries to be an integer or None'
+                )
+
+            if cached.off or max_cache_entries == 0:
                 return func(*args, **kwargs)
 
             keys = args
@@ -73,8 +73,8 @@ def function_cache(expected_max_entries=1000):
                 # # something is wrong if we are here more than expected
                 # # empty and turn it off
                 if (
-                    expected_max_entries is not None and
-                    cached.missed > expected_max_entries
+                    max_cache_entries is not None and
+                    cached.missed > max_cache_entries
                 ):
                     cached.off = True
                     cached.cache.clear()
--- a/premailer/cache.py
+++ b/premailer/cache.py
@@ -43,7 +43,7 @@ def function_cache():
                 not isinstance(max_cache_entries, int)
             ):
                 raise TypeError(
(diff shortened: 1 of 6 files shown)


premailer/cache.py
```diff
@@ -1,83 +1,61 @@
...
-# if we only have nonlocal
-class _Cache(object):
-    def __init__(self):
-        self.off = False
-        self.missed = 0
-        self.cache = {}
-
-
-def function_cache(expected_max_entries=1000):
+def function_cache():
```
peterbe: Now is not the time but the correct name ought to be "function_memoize" or something like that.

premailer/cache.py
```diff
@@ -1,83 +1,61 @@
...
-                # # empty and turn it off
-                if (
-                    expected_max_entries is not None and
-                    cached.missed > expected_max_entries
-                ):
-                    cached.off = True
-                    cached.cache.clear()
+
+                if max_cache_entries is None or len(cache) < max_cache_entries:
+                    cache[hashed] = result
```
peterbe: It's a bit strange that the caching just decides to stubbornly refuse to add more entries. I've already forgotten what you said in the justification in the PR description. Can you please add a code comment here to explain this reasoning.

premailer/premailer.py
```diff
@@ -132,33 +167,64 @@ def __init__(self, html, base_url=None,
                  disable_basic_attributes=None,
                  disable_validation=False,
                  cache_css_parsing=True,
+                 cache_css_parsing_size=1000,
+                 cache_css_output=True,
+                 cache_css_output_size=1000,
```
peterbe: I don't think these need to be here at all. They're operational, not functional. They clutter up the code and obscures the options. 

The likelihood of someone preferring to only set it to 500 or 10_000 is extremely rare and I think there are much better ways to do that. Like this:

```python
import os

def function_cache():
    ...snip...
    # indicator of cache missed
    sentinel = object()

    max_cache_entries = int(os.environ.get('PREMAILER_MAX_CACHE_ENTRIES', 1000))

    @functools.wraps(func)
    def inner(*args, **kwargs):
        ...
    return inner
```

That will make the code a *lot* simpler. It'll give all the power to your use case too

premailer/premailer.py
```diff
@@ -132,33 +167,64 @@ def __init__(self, html, base_url=None,
...
+        cache_css_parsing: Specifies whether to cache the CSS parsing results.
+
+        cache_css_parsing_size: Specifies the size for various CSS parsing
+            caches. If set to None, the size of the caches will not be limited.
+
+        cache_css_output: Specifies whether to cache the CSS output results.
+
+        cache_css_output_size: Specifies the size for various CSS output
+            caches. If set to None, the size of the caches will not be limited.
+        '''
```
peterbe: AWESOME!!

premailer/premailer.py
```diff
@@ -132,33 +167,64 @@ def __init__(self, html, base_url=None,
...
+
+        include_star_selectors: Specifies whether to process or ignore
+            selectors like '* { foo:bar; }'
+
+        cache_css_parsing: Specifies whether to cache the CSS parsing results.
+
+        cache_css_parsing_size: Specifies the size for various CSS parsing
+            caches. If set to None, the size of the caches will not be limited.
+
+        cache_css_output: Specifies whether to cache the CSS output results.
```
peterbe: Unlike the `cache_*_size` options, this one is actually *not* operational. Let's keep this. Setting this *will* potentially have an effect on the generated HTML output.

peterbe:
Actually, the more I think about it, we should consider CacheTools. It's just so convenient and we don't need to do *any* heavy lifting mess with the difference between Python 2 and 3. 

I still think the max cache size should be controlled from a documented environment variable though.

peterbe:
By the way, see https://github.com/peterbe/premailer/pull/203 
Watch out for merge conflicts if you rebase this.

peterbe:
cachetools is awesome. I think we should continue this effort over in https://github.com/peterbe/premailer/pull/206

peterbe:
https://github.com/peterbe/premailer/pull/206 landed.

PR #199, opened 2018-10-04
Minor README update

onecrayon:
Per discussion in #197, this PR includes some minor improvements to the README:

* FIXED: Initial sample code now runs on Python 3 (thanks to `print()` syntax)
* FIXED: Initial sample code now shows accurate output via latest premailer transform logic
* ADDED: Placeholder value for `base_url`
* ADDED: Example block with `data-premailer="ignore"` to initial sample code
* ADDED: Explicit documentation on the defaults used by `transform` function (also serves as basic documentation for the Premailer class options)

I'm not sure about the formatting of the documentation for default options for `transform`; might deserve its own section with a proper list instead of a code snippet. I'm open to suggestions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,7 @@ Next, the most basic use is to use the shortcut function, like this:
     ...         <h1 style="font-weight:bolder">Peter</h1>
     ...         <p>Hej</p>
     ...         </html>
-    ... """, base_url=None))
+    ... """))
     <html>
     <head>
         <style type="text/css">p::first-letter {float:left}</style>
@@ -102,6 +102,7 @@ The ``transform`` shortcut function transforms the given HTML using the defaults
 
 ::
 
+    base_url=None, # Optional URL prepended to all relative links (both stylesheets and internal)
     disable_link_rewrites=False, # Allow link rewrites (e.g. using base_url)
     preserve_internal_links=False, # Do not preserve links to named anchors when using base_url
     preserve_inline_attachments=True, # Preserve links with cid: scheme when base_url is specified
--- a/README.rst
+++ b/README.rst
@@ -98,7 +98,7 @@ Next, the most basic use is to use the shortcut function, like this:
     </body>
     </html>
 
-The `transform` shortcut method transforms the given HTML using the defaults for all options:
+The ``transform`` shortcut function transforms the given HTML using the defaults for all options:
 
 ::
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -71,25 +71,59 @@ Next, the most basic use is to use the shortcut function, like this:
...
+        </style>
+    </head>
     <body>
-        <h1 style="font-weight:bolder; border:1px solid black">Peter</h1>
+        <h1 style="border:1px solid black; font-weight:bolder">Peter</h1>
         <p style="color:red">Hej</p>
     </body>
     </html>
 
+The `transform` shortcut method transforms the given HTML using the defaults for all options:
```
peterbe: That should be double backticks. This is ReST after all :)

README.rst
```diff
@@ -71,25 +71,59 @@ Next, the most basic use is to use the shortcut function, like this:
...
     ...         p::first-letter { float:left; }
     ...         </style>
+    ...         <style type="text/css" data-premailer="ignore">
+    ...         h1 { color:blue; }
+    ...         </style>
     ...         <h1 style="font-weight:bolder">Peter</h1>
     ...         <p>Hej</p>
     ...         </html>
-    ... """)
+    ... """, base_url=None))
```
peterbe: Why did you put this one in?

README.rst
```diff
@@ -71,25 +71,59 @@ Next, the most basic use is to use the shortcut function, like this:
...
     >>> from premailer import transform
-    >>> print transform("""
+    >>> print(transform("""
     ...         <html>
     ...         <style type="text/css">
     ...         h1 { border:1px solid black }
     ...         p { color:red;}
     ...         p::first-letter { float:left; }
     ...         </style>
+    ...         <style type="text/css" data-premailer="ignore">
```
peterbe: Excellent!

README.rst
peterbe: It's not a "shortcut method" it's a "shortcut function".

README.rst
onecrayon: Idea here was to signal people that `transform` accepts this keyword argument in case they didn't make it further down (similar to the inclusion of `data-premailer="ignore"`). I'm not married to the idea if you don't like it, though. :-)

Incidentally (and a little off topic) is there a reason that `transform` doesn't just accept `**kwargs` and then pass it along to the Premailer constructor? I was surprised it supported `base_url` but none of the other options.

peterbe: Wow! It doesn't?! That's nuts. Can you fix it please?

The only reason the `transform` function exists at all is because it looks slightly simpler than:

```python
instance = Premailer(html)
output = instance.transform()
```

PR #206, opened 2018-10-09
use cachetools for caching

peterbe:
Hey @p12tic I feel bad for "shooting you down" in https://github.com/peterbe/premailer/pull/198

This PR is a good start and it's taking a bunch of ideas from your PR but it's also a LOT easier. 

The code here is NOT finished. Can you please do the rest of the work?

- [ ] The [original PR](https://github.com/peterbe/premailer/pull/198/files) had more tricks to break out functions so that they can have a cache decorator. E.g. `_css_rules_to_string`
- [x] The environment variable `PREMAILER_CACHE_MAXSIZE` needs to be documented in the readme. 
- [ ] Put some more, heavier, blocks of HTML and CSS in the `stresstest/samples/` directory. 
- [ ] A `tox.ini` that makes is possible to run the test suite for 2.7 and 3.x in locally and the make travis execute that same `tox.ini`.


peterbe:
Ping @p12tic 

Any chance you can look into this one?

p12tic:
I've reconsidered and think that the performance of the library is good enough given the additional time investment needed to bring in the changes.

peterbe:
This PR isn't perfect but it's a huge improvement. `cachetools` is a great library and this PR removes a bunch of ad-hoc code in favor of `cachetools`.

Issue #207, opened 2018-11-13
Premailer tries download css 


TZanke:
Hello,

in our customer environments, internet access is not allowed, but Premailer tries to download css files and encounters an exception. I could no find any option to disable this feature.

```
File "premailer/premailer.py", line 341, in transform  
    css_body = self._load_external(href)

File "premailer/premailer.py", line 520, in _load_external
    css_body = self._load_external_url(url)

File "premailer/premailer.py", line 507, in _load_external_url
    return requests.get(url).text

File "requests/api.py", line 70, in get
    return request('get', url, params=params, **kwargs)

File "requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)

File "requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)

File "requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)

File "requests/adapters.py", line 487, in send
    raise ConnectionError(e, request=request)

ConnectionError: HTTPSConnectionPool(host='fonts.googleapis.com', port=443): Max retries exceeded with url: /css?family=Open+Sans:300,400,600,700 (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7fe912dcec10>: Failed to establish a new connection: [Errno 101] Das Netzwerk ist nicht erreichbar',))
```

peterbe:
Is it possible that you have something like this:

```html
<html>
<head>
<link rel="stylesheet" href="https://fonts.googleapis.com?family=Lato">
<style>
css { key: value }
</style>
<body>
  ...
</body>
</html>
```

In that case, try setting the `data-premailer="ignore"` attribute onto that `<link>` tag.

TZanke:
This is correct. Cause we are using Premailer to generate the reply Blockquote, the source mail is from external sources. Adding a ignore attribute comes at same cost as removing the entire tag.

peterbe:
But if you put an ignore tag on the font stylesheet, doesn't that solve your problem? Or do you have a stylesheet that uses `@import externalURL`?

TZanke:
`data-premailer="ignore"` works as expected. But in both cases i have to manipulate external first. Would be nice to have a Premailer parameter which disables external loading at all.

peterbe:
Ah! Like an option that says "Don't ever bother doing any more downloads when you process this."
[...]

PR #214, opened 2019-01-18
add skip_load_external

TZanke:
Fixes https://github.com/peterbe/premailer/issues/207

Please review my code again.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,11 +61,9 @@ premailer does this. It parses an HTML page, looks up ``style`` blocks
 and parses the CSS. It then uses the ``lxml.html`` parser to modify the
 DOM tree of the page accordingly.
 
-::
-
-    Warning!
-    By default, premailer will attempt to download any external stylesheets by URL over the Internet.
-    If you want to prevent this you can use the ``allow_network=False`` option.
+Warning!
+By default, premailer will attempt to download any external stylesheets by URL over the Internet.
+If you want to prevent this you can use the ``allow_network=False`` option.
 
 Getting started
 ---------------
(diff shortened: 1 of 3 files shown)


premailer/premailer.py
```diff
@@ -540,6 +544,8 @@ def _load_external_url(self, url):
     def _load_external(self, url):
         """loads an external stylesheet from a remote url or local path
         """
+        if self.skip_load_external:
```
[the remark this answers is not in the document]
peterbe: Since it's not many places where `self._load_external()` is called, I think you can be confident it won't get here if `skip_load_external` *is* `False`.

TZanke:
I implemented your suggestions and renamed parameter to `allow_network` which defaults to True. But i would like to set it to False, because in corporate environments we prefer security related settings to be as secure as possible by default. At the moment, without this branch, i ended up in ExternalNotFoundError, because network access is limited. Without our firewalls, premailer would download unknown files to our systems without notice.

peterbe:
Great work! I like it. Just one last thing; can you add a unit test that uses `allow_network=False`?

Also, regarding the default, I doubt it's feasible to change the default to `False` because it's a breaking change and it's rather strange that a server isn't allowed to reach for things on the Internet. Although I work in a security paranoid enterprise myself (Mozilla), I've never heard of this concern. But I don't dispute it. 

Perhaps you can add a blurb to the README that says "Warning! By default, `premailer` will attempt to download any external stylesheets by URL over the Internet. If you want to prevent this you can use the `allow_network=False` option."

peterbe:
Ping for a unit test added to the PR.

README.rst
```diff
@@ -61,6 +61,12 @@ premailer does this. It parses an HTML page, looks up ``style`` blocks
 and parses the CSS. It then uses the ``lxml.html`` parser to modify the
 DOM tree of the page accordingly.
 
+::
+
+    Warning!
+    By default, premailer will attempt to download any external stylesheets by URL over the Internet.
+    If you want to prevent this you can use the allow_network=False option.
```
peterbe:
```suggestion
    If you want to prevent this you can use the `allow_network=False` option.
```

peterbe:
```suggestion
    If you want to prevent this you can use the ``allow_network=False`` option.
```

PR #222, opened 2019-04-25
Set exclude_pseudoclasses to False by default to have the same behavi…

calixteman:
…our as the command line

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -155,7 +155,7 @@ module.
     --remove-internal-links PRESERVE_INTERNAL_LINKS
                           Remove links that start with a '#' like anchors.
     --exclude-pseudoclasses
-                          Pseudo classes like p:last-child', p:first-child, etc
+                          Pseudo classes like p:last-child', p:first-child, :nth-child, etc
     --preserve-style-tags
                           Do not delete <style></style> tags from the html
                           document.
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -140,7 +140,7 @@ class Premailer(object):
         disable_link_rewrites=False,
         preserve_internal_links=False,
         preserve_inline_attachments=True,
-        exclude_pseudoclasses=False,
+        exclude_pseudoclasses=True,
         keep_style_tags=False,
         include_star_selectors=False,
         remove_classes=False,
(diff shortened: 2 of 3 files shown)


premailer/premailer.py
```diff
@@ -140,7 +140,7 @@ def __init__(
         disable_link_rewrites=False,
         preserve_internal_links=False,
         preserve_inline_attachments=True,
-        exclude_pseudoclasses=True,
+        exclude_pseudoclasses=False,
```
peterbe: This makes me nervous. Most people use premailer programmatically rather that the `__main__.py` endpoint. A change here is expected to break the tests. (At the time of writing I'm waiting for Travis to finish) And potentially it might change peoples implementations when they upgrade to the next version. 
Which is right I'm not sure about but I'm sure we should have the same defaults in `__main__`, this file and in the README.

peterbe:
Seems @coveralls is broken. "Coverage remained the same at ?% when "

calixteman:
So finally, only the doc changed to avoid to break something.
So we'll have at least something to grep.

README.rst
```diff
@@ -155,7 +155,7 @@ module.
     --remove-internal-links PRESERVE_INTERNAL_LINKS
                           Remove links that start with a '#' like anchors.
     --exclude-pseudoclasses
-                          Pseudo classes like p:last-child', p:first-child, etc
+                          Pseudo classes like p:last-child', p:first-child, :nth-child, etc
```
peterbe: Can you make this change in `__main__.py` too? Line ~53.

peterbe:
You know what sucks? With `argparse` there is no way to NOT exclude pseudoclasses. 
All the boolean'y options in `argparse` needs to be `False` by default.

README.rst
calixteman: I already did it.

calixteman:
Or we could store_false instead but that means that we must rename the option and so... break something, somewhere

peterbe:
Wait! What am I saying! If you use `python -m premailer` then you get `exclude_pseudoclasses False` but if you use `premailer.transform(html)` you get `exclude_pseudoclasses True`. Sigh.

peterbe:
oops.

peterbe:
For now, can you just make that one last extra edit to the README (about `:nth-child`) and then we'll worry about differences in `__main__` vs. programmatically later. 
Also, can you rebase to edit the commit message.

Issue #225, opened 2019-06-23
Thread synchronisation issue during cache access.


asandeep:
Looks like I got hit with a thread synchronisation issue in Premailer's cache. Here's my usecase:

I am parsing thousands of HTML documents every hour using Premailer. Now each document can have its own set of CSS defined inline. Apart from that, I have a fixed CSS file that I apply to every document that gets processed. To speed things up, I start multiple thread workers that parses the document in parallel.

The program runs fine till the time Premailer's cache gets completely filled, after which it starts removing least recently used items. Now since multiple threads are accessing same cache there are chances that a rule removed by one thread was being access by another. Below is the traceback of exception that is raised in this case:

`File "/home/ubuntu/Documents/setups/anaconda2/envs/sentieo/lib/python2.7/site-packages/premailer/premailer.py", line 442, in transform
   sel = _create_cssselector(selector)
 File "/home/ubuntu/Documents/setups/anaconda2/envs/sentieo/lib/python2.7/site-packages/cachetools/__init__.py", line 48, in wrapper
   cache[k] = v
 File "/home/ubuntu/Documents/setups/anaconda2/envs/sentieo/lib/python2.7/site-packages/cachetools/lru.py", line 21, in __setitem__
   cache_setitem(self, key, value)
 File "/home/ubuntu/Documents/setups/anaconda2/envs/sentieo/lib/python2.7/site-packages/cachetools/cache.py", line 52, in __setitem__
   self.popitem()
 File "/home/ubuntu/Documents/setups/anaconda2/envs/sentieo/lib/python2.7/site-packages/cachetools/lru.py", line 35, in popitem
   return (key, self.pop(key))
 File "/home/ubuntu/Documents/setups/anaconda2/envs/sentieo/lib/python2.7/site-packages/cachetools/abc.py", line 39, in pop
   raise KeyError(key)
KeyError: (u'.separator',)`

The cachetools documentation already states that Cache implementation classes are not thread-safe (See note section here: https://cachetools.readthedocs.io/en/latest/#cache-implementations). But it is possible to synchronise thread by passing a lock instance to cachetools memoizing decorators (https://cachetools.readthedocs.io/en/latest/#cachetools.cached).

For now I worked my way around this by increasing the cache size and hacking the cachetools.LRUCache implementation to return LFUCache instead (code below) which works in my case for now but this just delays the disaster and not the full-proof solution.

```python
def _hacked_lfu_cache(maxsize):
    """
[...]

PR #227, opened 2019-07-06
Use lock to synchronize multiple thread access to Premailer cache and added environment variables to choose a different cache implementation.

asandeep:


README.rst
```diff
@@ -299,6 +299,19 @@ this, you have to pass in ``cssutils_logging_handler`` and
...
+Choosing the cache implementation
+-----------------------------------
+
+By default, ``premailer`` uses `LFUCache
+<https://cachetools.readthedocs.io/en/latest/#cachetools.LRUCache>`__ to cache
+CSS styles during parsing. But it is possible to choose a different
+implementation using below environment variables:
+
+- ``PREMAILER_CACHE``: Can be LRU, LFU, RR or TTL. Default is LRU.
+- ``PREMAILER_CACHE_MAXSIZE``: Maximum no. of items to be stored in cache.
```
peterbe: Should mention what the default is (or link to cachetools page that specifies that).

README.rst
```diff
@@ -299,6 +299,19 @@ this, you have to pass in ``cssutils_logging_handler`` and
     >>> mylog.getvalue()
     'CSSStylesheet: Unknown @rule found. [2:1: @keyframes]\n'
 
+Choosing the cache implementation
+-----------------------------------
+
+By default, ``premailer`` uses `LFUCache
+<https://cachetools.readthedocs.io/en/latest/#cachetools.LRUCache>`__ to cache
```
peterbe: LFUCache or LRUCache?

peterbe: Ie. which is right, the link href or the link text?

README.rst
```diff
@@ -299,6 +299,19 @@ this, you have to pass in ``cssutils_logging_handler`` and
...
+CSS styles during parsing. But it is possible to choose a different
```
peterbe: It's not just for "CSS styles during parsing", right?

README.rst
```diff
@@ -299,6 +299,19 @@ this, you have to pass in ``cssutils_logging_handler`` and
...
+implementation using below environment variables:
+
+- ``PREMAILER_CACHE``: Can be LRU, LFU, RR or TTL. Default is LRU.
+- ``PREMAILER_CACHE_MAXSIZE``: Maximum no. of items to be stored in cache.
+- ``PREMAILER_CACHE_TTL``: Time to live for cache entries. Only applicable for TTL cache.
+
+
```
peterbe: It should mention **why** you'd want to mess with these things. That's an early question that's going to come up in minds. 

Perhaps you can't contribute an explanation but if so it should be mentioned something like "These are advanced options that probably don't matter for most people with regular load."

Anything to put peoples minds' at ease.

premailer/cache.py
```diff
@@ -1,15 +1,43 @@
...
+
+# Time to live (seconds) for entries in TTL cache. Defaults to 1 hour.
+TTL_CACHE_TIMEOUT = 1 * 60 * 60
+
+# Lock to prevent multiple threads from accessing the cache at same time.
+CACHE_ACCESS_LOCK = threading.RLock()
+
+cache_type = os.environ.get("PREMAILER_CACHE", "LRU")
+if cache_type not in CACHE_IMPLEMENTATIONS:
+    raise Exception(
```
peterbe:
```suggestion
    raise ValueError(
```

README.rst
peterbe: Judging by the code it's LRUCache

premailer/cache.py
```diff
@@ -1,15 +1,43 @@
...
+import cachetools
 
 
-cache = LRUCache(maxsize=int(os.environ.get("PREMAILER_CACHE_MAXSIZE", 128)))
+# Available cache options.
+CACHE_IMPLEMENTATIONS = {
+    "LFU": cachetools.LFUCache,
+    "LRU": cachetools.LRUCache,
+    "RR": cachetools.RRCache,
+    "TTL": cachetools.TTLCache,
```
peterbe: These two; why would you ever want them? Seems like terrible choices. 

Actually, more importantly: Why even give the user the choice to switch to LFU cache? How does it differ from LRU for users? Or does it even matter? If it matters, why isn't one option best for everyone. (thinking about simplicity)

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,52 @@
...
+                del os.environ[key]
+            except KeyError:
+                pass
+
+    def test_unknown_cache(self):
+        os.environ["PREMAILER_CACHE"] = "UNKNOWN"
+
+        with self.assertRaises(Exception) as assert_context:
+            cache_fd = open("premailer/cache.py", "r")
+            imp.load_source("cache.py", "premailer/cache.py", cache_fd)
```
peterbe: Will this leave an open file descriptor here? Ever? Why not:
```python
with open("premailer/cache.py") as fd:
    imp.load_source("cache.py", "premailer/cache.py", fd)
```

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,52 @@
...
+            try:
+                del os.environ[key]
+            except KeyError:
+                pass
+
+    def test_unknown_cache(self):
+        os.environ["PREMAILER_CACHE"] = "UNKNOWN"
+
+        with self.assertRaises(Exception) as assert_context:
+            cache_fd = open("premailer/cache.py", "r")
```
peterbe: The use of forward slashes, will that not break developers on Windows? I.e. use `os.path.join('premailer', 'cache.py')` instead.

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.assertTrue(
+            assert_context.exception.args[0].startswith(
+                "Unsupported cache implementation"
+            )
+        )
+
+    def test_default_cache(self):
+        cache_fd = open("premailer/cache.py", "r")
+        cache_module = imp.load_source("cache.py", "premailer/cache.py", cache_fd)
```
peterbe: I'm just uncomfortable that this won't be left open for the duration of the whole test suite. Or will `cache_fd.close` always be called at the end of the test method?

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,52 @@
...
+            cache_fd = open("premailer/cache.py", "r")
+            imp.load_source("cache.py", "premailer/cache.py", cache_fd)
+
+        self.assertTrue(
+            assert_context.exception.args[0].startswith(
+                "Unsupported cache implementation"
+            )
+        )
+
+    def test_default_cache(self):
```
peterbe: I don't know if this test is really necessary. It's a deep implementation detail. The tests should really be more about the "user API" not about how things are done internally. The contents of this test method doesn't really ever mimic something a user would do.

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,52 @@
...
+        os.environ["PREMAILER_CACHE"] = "TTL"
+        os.environ["PREMAILER_CACHE_TTL"] = "10"
+        os.environ["PREMAILER_CACHE_MAXSIZE"] = "50"
+
+        cache_fd = open("premailer/cache.py", "r")
+        cache_module = imp.load_source("cache.py", "premailer/cache.py", cache_fd)
+
+        self.assertEquals(type(cache_module.cache), cachetools.TTLCache)
+        self.assertEquals(cache_module.cache.maxsize, 50)
+        self.assertEquals(cache_module.cache.ttl, 10)
```
peterbe: The most important change of this PR is that it enabled a thread locking mechanism but there's no unit test to lock that in (no pun intended). 

Can you, first comment out the `threading.RLock()` stuff above, then create a test that starts two threads and somehow trigger the race condition. Then uncomment the threading stuff and check that it solved the problem.

premailer/cache.py
```diff
@@ -1,15 +1,43 @@
...
+}
+
+# Time to live (seconds) for entries in TTL cache. Defaults to 1 hour.
+TTL_CACHE_TIMEOUT = 1 * 60 * 60
+
+# Lock to prevent multiple threads from accessing the cache at same time.
+CACHE_ACCESS_LOCK = threading.RLock()
```
peterbe: This should be lowercase since it's not really a human chosen constant and it's never really configurable. Note that it's called `cache` on line 35 (not `CACHE`).

README.rst
asandeep: Updated doc with default values for each option.

README.rst
asandeep: Its LRU. Sorry for the confusion.

README.rst
asandeep: Yes right. Updated.

README.rst
asandeep: I agree. These options are meant for advanced usage. Created a separate section  for advanced configuration and moved the section there.

premailer/cache.py
asandeep: Makes sense. Done.

premailer/cache.py
asandeep: I mentioned the usecase where LFU is more suitable than LRU here: https://github.com/peterbe/premailer/issues/225.
I think TTL can still be used in scenarios where after initial parsing, only subset of cache entries are used repetedly, leaving behind entries that are just sitting to consume the memory. Here, TTL can help to clean up those entries after a certain time period.
RR, I agree, doesn't look like it is really useful in this case. Removed it from list of available choices.

premailer/tests/test_cache.py
asandeep: Nice catch. Thanks for pointing it out.
I actually don't even need it. Removed.

premailer/tests/test_cache.py
asandeep: Totally makes sense. Updated.

premailer/tests/test_cache.py
asandeep: I agree with you. Removed.

premailer/tests/test_cache.py
asandeep: Its really hard to reproduce this race condition with sureity without modifying the cachetools code. I was actually able to reproduce it using below code, but for that I had to modify cachetools code to insert some delay while poping items from cache here: https://github.com/tkem/cachetools/blob/master/cachetools/lru.py#L35

```python
os.environ["PREMAILER_CACHE_MAXSIZE"] = "1"

from premailer import cache

RULES_MAP = {
    "h1": "{ color: blue; }",
    "h2": "{ color: green; }",
}


@cache.function_cache()
def get_styles(rule):
    return RULES_MAP[rule]


class TestFunctionCache(unittest.TestCase):
    def test_cache_multithread_synchronization(self):
        class RuleMapper(threading.Thread):
            def run(self):
	        for rule in RULES_MAP:
                    get_styles(rule)

        threads = []
        for _ in range(2):
            threads.append(RuleMapper())

        for thread in threads:
            thread.start()

        for thread in threads:
            thread.join()
```
Below are links to couple of issues that talks about this scenario if that works.
https://github.com/tkem/cachetools/issues/80
https://github.com/tkem/cachetools/issues/82

premailer/tests/test_cache.py
asandeep: > Why is the file called `.red`?
This was added by mistake. Removed now.

asandeep:
> I think the most important thing was the use of `threading.RLock()` but the PR mixes in a bunch of other things that I don't see to be relevant. Sorry. But great work on that too by the way!!
> 
> I don't yet see the point of switching from LRU to LFU or even giving the option. It just feels like more moving parts that adds noise for no _great_ reason.

For the usecase I mentioned in the issue: https://github.com/peterbe/premailer/issues/225, LFU makes more sense than LRU but I agree that such scenarios are not very common and are applicable for advanced usage only. I have updated the documentation to mention the same.
If you still feel this won't be a value add, let me know and I would be happy to remove it.

premailer/tests/test_cache.py
peterbe: I have no idea how you would do it but perhaps some inspiration here: 
https://github.com/peterbe/premailer/blob/4d74656fb12e8e44683fa787ae71c0735282376b/premailer/tests/test_premailer.py#L1846

I didn't actually write that code. It was a contribution. But it did (sometimes (I think)) triggered the thread-un-safety and since the contribution that test no longer fails any more.

premailer/cache.py
peterbe: I see that you switch from `LRUCache({max_size: 128})` to `LFUCache({max_size: 1000})` but that's changing to things at once. Being able to switch from 128 to 1000 was already possible with setting `PREMAILER_CACHE_MAXSIZE` without the need for your hack. 
But why change from LRU to LFU?
And if you've done the research and found that LFU is much better when you're processing thousands of HTML documents, why wouldn't everyone prefer LFU?

premailer/tests/test_cache.py
asandeep: Thanks for the link. I was actually able to add the test by providing a overridden implementation of cache that adds some delay during delete operation on cache, thus resulting in race condition.
The test will fail if no lock is passed to `cached` decorator.

asandeep:
> Sorry for sounding difficult and resistant. I'm just wondering if there is a simpler solution to it all.
> 
> In particular, I'd love to see a piece of code that reproduces the problem. Next I'd love to see why it's not enough to just increase the `maxsize` and if LFU is better than LRU then perhaps we should just hardcode that for everyone.

Sure I understand. Let me try to provide a better explanation.

Suitability of a cache implementation depends on the usage scenario. So, basically `LRU` will keep most recently used items, while discarding the rest. On the other hand, `LFU` keeps items that are most frequently used. Now the frequently used items may or may not be recently used as well. Putting this into context, consider below scenarios:

*Scenario 1*:
A set of common rules are defined at top of every document.
Individual documents can also define rules that are specific to those documents.
Now when premailer starts processing it will cache the common rules first and then the specific rules.
The next document will need same set of common rules that were added to cache when processing first document.
In this scenario, `LFU` will preserve the common rules in cache while `LRU` would have discarded the same.

*Scenario 2*:
A set of unrelated documents are processed using `premailer` where every document defines its own set of rules.
In this scenario, `LRU` would be more suitable as it will preserve rules for latest document being processed discarding the older ones.

So, which cache implementation is better really depends on the usage scenario. 
I can certainly write the tests but they will be more about testing internal workings of different cache implementations.

Hope this makes sense. Let me know what you think.

peterbe:
Suppose you have these documents, and a max_size of 3:

```html
<style> p { color: red; }</style>
<style> div { margin: 0; }</style>
<style> div { margin: 10px; }</style>
```

```html
<style> p { color: red; }</style>
<style> div { margin: 0; }</style>
<style> div { margin: 20px; }</style>
```

```html
<style> p { color: red; }</style>
<style> div { margin: 0; }</style>
<style> div { margin: 30px; }</style>
```

Then, LFU will be better. As it encounters document number 2 it knows to kick out the least recently used one and keep the frequently reoccurring. 

That's a realistic scenario for premailer, isn't it?

In Scenario 2; if the documents are all different then no caching can be achieved, right? So it'd be more interesting to **disable** caching to avoid unnecessary memory usage that will all be useless. 

What I'm alluding to is; if you think Scenario 1 (as you and I describe) it is very realistic for people who turn *many* HTML into email-HTML then LFU should be the default and only default. 

But at the same time, now the code is written so we might as well keep it.

peterbe:
If your contribution solves the locking problem at caching level, does that mean these lines can be removed: https://github.com/peterbe/premailer/blob/03e8848a3de8118e127b7077ae35c25cb843c496/premailer/merge_style.py#L30

After all, if you have two concurrent threads calling this same code:

```python
@basic_cache_decorator()
def an_thread_unsafe_function(arg):
    return calling_some_thread_unsafe_library(arg)
```

then, if they're sufficiently concurrent, there's a chance that `calling_some_thread_unsafe_library(arg)` gets executed by each thread concurrently!. So you can solve it like this:

```python
@basic_cache_decorator()
def an_thread_unsafe_function(arg):
    with lock:
        return calling_some_thread_unsafe_library(arg)
lock = threading.RLock()
```
But then, each thread might *each* call `calling_some_thread_unsafe_library(arg)` *synchronously* which might be bad for performance. 

So, with your solution, using `cachetools`:

```python
@cachetools.LFUCache(lock=lock)
def an_thread_unsafe_function(arg):
    return calling_some_thread_unsafe_library(arg)
lock = threading.RLock()
```
Then, the `calling_some_thread_unsafe_library(arg)` will only ever be called *once*. RIght? That second thread that is a millisecond slower would wait for the cache to fill up first. Right?

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+
+        threads = [RuleMapper() for _ in range(2)]
+        for thread in threads:
+            thread.start()
+
+        for thread in threads:
+            thread.join()
+
+        exceptions = [thread.exception for thread in threads if thread.exception]
+        self.assertEquals(
```
peterbe: I think this line(s) can just be simplified to:
```python
self.assertTrue(not exceptions)
```

premailer/tests/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self.assertEquals(type(cache_module.cache), cachetools.TTLCache)
+        self.assertEquals(cache_module.cache.maxsize, 50)
+        self.assertEquals(cache_module.cache.ttl, 10)
+
+    def test_cache_multithread_synchronization(self):
+        """
+        Tests thread safety of internal cache access.
+
+        The test will fail if function_cache would not have been thread safe.
```
peterbe: Great comment!

README.rst
```diff
@@ -334,6 +335,26 @@ environment variable to control is ``PREMAILER_CACHE_MAXSIZE``. This parameter
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, ``premailer`` uses `LRUCache
+<https://cachetools.readthedocs.io/en/latest/#cachetools.LRUCache>`__ to cache
+selectors, styles and parsed CSS strings. If LRU doesn't serve your purpose, it
+is possible to switch to an alternate implementation using below environment
+variables.
+
+- ``PREMAILER_CACHE``: Can be LRU, LFU or TTL. Default is LRU.
+- ``PREMAILER_CACHE_MAXSIZE``: Maximum no. of items to be stored in cache. Defaults to 128.
```
peterbe: I wonder, that 128 comes from https://docs.python.org/3/library/functools.html#functools.lru_cache I think. Who has a HTML document, with CSS that has 128 or less selectors?! Wouldn't it make sense to bump this up to at least 1,000?

Another option is to set it to `None` but I think people would be very negatively surprised if their memory explodes. 


By the way, they say, for `functools.lru_cache`:

> The LRU feature performs best when maxsize is a power-of-two.

I'm not sure how that applies to `cachetools`, can you find out, but then the better default would be 1024.

premailer/tests/test_cache.py
asandeep: Done

premailer/tests/test_cache.py
asandeep: Thanks :)

asandeep:
> Suppose you have these documents, and a max_size of 3:
> 
> ```
> <style> p { color: red; }</style>
> <style> div { margin: 0; }</style>
> <style> div { margin: 10px; }</style>
> ```
> 
> ```
> <style> p { color: red; }</style>
> <style> div { margin: 0; }</style>
> <style> div { margin: 20px; }</style>
> ```
> 
> ```
> <style> p { color: red; }</style>
> <style> div { margin: 0; }</style>
> <style> div { margin: 30px; }</style>
> ```
> 
> Then, LFU will be better. As it encounters document number 2 it knows to kick out the least recently used one and keep the frequently reoccurring.
> 
> That's a realistic scenario for premailer, isn't it?
> 
> In Scenario 2; if the documents are all different then no caching can be achieved, right? So it'd be more interesting to **disable** caching to avoid unnecessary memory usage that will all be useless.
> 
> What I'm alluding to is; if you think Scenario 1 (as you and I describe) it is very realistic for people who turn _many_ HTML into email-HTML then LFU should be the default and only default.
> 
> But at the same time, now the code is written so we might as well keep it.

You are right. LFU seems to be a more preferable choice compared to LRU. Updated the code accordingly.

asandeep:
> If your contribution solves the locking problem at caching level, does that mean these lines can be removed:
> 
> https://github.com/peterbe/premailer/blob/03e8848a3de8118e127b7077ae35c25cb843c496/premailer/merge_style.py#L30
> 
> After all, if you have two concurrent threads calling this same code:
> 
> ```python
> @basic_cache_decorator()
> def an_thread_unsafe_function(arg):
>     return calling_some_thread_unsafe_library(arg)
> ```
> 
> then, if they're sufficiently concurrent, there's a chance that `calling_some_thread_unsafe_library(arg)` gets executed by each thread concurrently!. So you can solve it like this:
> 
> ```python
> @basic_cache_decorator()
> def an_thread_unsafe_function(arg):
>     with lock:
>         return calling_some_thread_unsafe_library(arg)
> lock = threading.RLock()
> ```
> 
> But then, each thread might _each_ call `calling_some_thread_unsafe_library(arg)` _synchronously_ which might be bad for performance.
> 
> So, with your solution, using `cachetools`:
> 
> ```python
> @cachetools.LFUCache(lock=lock)
> def an_thread_unsafe_function(arg):
>     return calling_some_thread_unsafe_library(arg)
> lock = threading.RLock()
> ```
> 
> Then, the `calling_some_thread_unsafe_library(arg)` will only ever be called _once_. RIght? That second thread that is a millisecond slower would wait for the cache to fill up first. Right?

It might not work as we expect. The **Note** section [in cachetools docs]( https://cachetools.readthedocs.io/en/latest/#cachetools.cached) tells that the lock only prevents cache access from multiple threads and underlying function should implement thread-safety on its own, and there are chances that this code [here](https://github.com/tkem/cachetools/blob/master/cachetools/__init__.py#L60) (body of `csstext_to_pairs` function in our case) might get executed by multiple threads at same time.

README.rst
asandeep: I looked into [LRU implementation](https://github.com/python/cpython/blob/master/Lib/functools.py#L469) and couldn't figure out why they suggest the `maxsize to be power of two`.

One drawback of keeping higher maxsize that I see is that cache is cleared less frequently even if items in cache are never used (unless you are using TTL cache ofcourse) and with `cache_css_parsing` turned on, the memory usage can go really high.

I would like to leave it to default of 128 and let users choose appropriate value if 128 is too less for them.

Let me know your thoughts.

peterbe: Fine by me. 

I suspect that many users of `premailer` use it sporadically. E.g. when a form is submitted, an email is sent. 

If you're using it for big bulk processing, like you, you come here and learn about the max_size option.

asandeep:
> Let's do it!
> 
> Now, will you promise to maintain this feature if it fails for people? :)

Sure. Would love to. :)

PR #229, opened 2019-07-16
setup.py: set an upper bound on dependencies version

JordanP:
This should make this library more robust in case of breaking changes in one of
its dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,13 @@ def find_version(*file_paths):
     raise RuntimeError("Unable to find version string.")
 
 
-install_requires = ["lxml", "cssselect", "cssutils", "requests", "cachetools"]
+install_requires = [
+    "lxml<4.0",
+    "cssselect<2.0",
+    "cssutils<2.0",
+    "requests<3.0",
+    "cachetools<4.0",
+]
 
 if sys.version_info >= (2, 6) and sys.version_info <= (2, 7):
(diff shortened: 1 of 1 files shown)


moggers87:
LXML 4.0 was released in 2017. I assume that was a typo and you meant to put 5.0?

JordanP:
@moggers87 yeah, I think it was a typo. 

Let's first agree or not, whether this project should set a some major version boundaries on its dependencies. What do you think ?

moggers87:
I'm generally against the idea, unless a particular project is known to introduce breaking changes on major versions (e.g. Django, Celery). While users might get caught out by some unexpected changes, that is something the user can work around until there's a new release of premailer (i.e. they can set `lxml<5.0` in their requirements file).

If we do limit dependencies in this way we might end up locking users into old versions of libraries that have missing features, bugs, and security issues. Users would then have to wait for a new premailer release.

tl;dr: unless we know better, I think it's best to leave pinning dependencies to users

JordanP:
I understand. 

More or more libraries are following semver where a major version bumps means a non backward compatible. API. Agreed that, users can adjust their dependencies and say set `lxml<5.0 in their requirements file` but in a average Python project you get dozens of dependencies so easily a hundred of indirect dependencies, not convenient. 

But it's a tradeoff and your stand is totally valid and shared by many too. .

PR #232, opened 2019-07-27
Add allow_insecure_ssl option for request.get() on external URLs

majortom731:
Added a new option and cmdline switch called 'allow_insecure_ssl' which, when enabled, will cause request.get() to be called with verify=False as additional parameter.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ never too late to start, so let's start here and now.
 dev
 -----
 
-* Add allow_insecure_ssl option for external URLs
+* Add ``allow_insecure_ssl`` option for external URLs
 
 3.5.0
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ premailer Changes
 Peter's note: Unfortunately, ``premailer`` has never kept a change log. But it's
 never too late to start, so let's start here and now.
 
+dev
+-----
+
+* Add allow_insecure_ssl option for external URLs
+
 3.5.0
 -----
 
--- a/README.rst
+++ b/README.rst
@@ -133,7 +133,7 @@ The ``transform`` shortcut function transforms the given HTML using the defaults
     align_floating_images=True, # Add align attribute for floated images
     remove_unset_properties=True # Remove CSS properties if their value is unset when merged
     allow_network=True # allow network access to fetch linked css files
-    allow_insecure_ssl=False # don't allow unverified SSL certificates for exterenal links
+    allow_insecure_ssl=False # Don't allow unverified SSL certificates for external links
 
 For more advanced options, check out the code of the ``Premailer`` class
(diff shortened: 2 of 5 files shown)


majortom731:
Hi! Glad you like it!

Well, insecure is what it is...The idea is taken from the curl commandline tool, where it is called "--insecure". But I can change it to "--skip-verify" if you like that better... 

Github's travis CI reports a failed build because the py34 test env uses an incompatible libxml or something. Is that important?

peterbe:
Any idea how to solve (in a separate issue or PR) the problem of the lxml install in tox?

majortom731:
Hmm... I found out [here](https://pypi.org/project/lxml/4.4.0/) that python 3.4 support was dropped in version 4.4.0 of lxml. So I tried to force tox to use the lxml version before 4.4.0 for the py34 env like this (in tox.ini):
`[testenv:py34]
deps =
    lxml<=4.3.5
`
Seems to work...

peterbe:
If you prefer the name `allow_insecure_ssl` I'm not going to block. Go for it. 

However, what you need to do is to add a mention of this on the README (in two places!)

Also, in the [`CHANGES.rst`](https://github.com/peterbe/premailer/blob/master/CHANGES.rst) can you add a new section with the headline called `dev`. Then, when I make a release next I'll replace the word `dev` with the next number. 

```
dev
-----

* Write your change here...
```

majortom731:
Had to fiddle around with git a little, but now I think it's ready, all in 1 commit.

The tox.ini change is in #233

peterbe:
You don't need to squash your commits. I'll do that when I merge. But you'll now need to rebase so the tests don't fail.

majortom731:
Sorry, I'm still new to this. :)
Now ok?

CHANGES.rst
```diff
@@ -4,6 +4,11 @@ premailer Changes
 Peter's note: Unfortunately, ``premailer`` has never kept a change log. But it's
 never too late to start, so let's start here and now.
 
+dev
+-----
+
+* Add allow_insecure_ssl option for external URLs
```
peterbe:
```suggestion
* Add ``allow_insecure_ssl`` option for external URLs
```

README.rst
```diff
@@ -133,6 +133,7 @@ The ``transform`` shortcut function transforms the given HTML using the defaults
     align_floating_images=True, # Add align attribute for floated images
     remove_unset_properties=True # Remove CSS properties if their value is unset when merged
     allow_network=True # allow network access to fetch linked css files
+    allow_insecure_ssl=False # don't allow unverified SSL certificates for exterenal links
```
peterbe:
```suggestion
    allow_insecure_ssl=False # Don't allow unverified SSL certificates for external links
```

peterbe:
I'll make a new release for this 
See https://github.com/peterbe/premailer/pull/234

PR #235, opened 2019-08-23
Pass validate to csstext_to_pairs

mbenedettini:
Currently `disable_validation` does not work as it should since `csstext_to_pairs` does not receive the right `validate` parameter in order to disable validation.

### DIFF
--- a/premailer/__init__.py
+++ b/premailer/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import, unicode_literals
 from .premailer import Premailer, transform  # noqa
 
-__version__ = "3.6.0"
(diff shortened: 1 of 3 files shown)


premailer/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import, unicode_literals
 from .premailer import Premailer, transform  # noqa
 
-__version__ = "3.6.0"
+__version__ = "3.6.1"
```
peterbe:
```suggestion
__version__ = "3.6.0"
```
I'll update that line when I make a release next.

PR #252, opened 2021-04-09
Add preserve_handlebar_syntax option

### LINKED ISSUE #248: Mustache {{ }} syntax in href is broken on transform
Hello

Doing a premailer pass causes anchor with template mustaches in href, such as 
`<a href="{{ unsubscribe_link }}">Unsubscribe</a>` 
to be broken like 
`<a href="%7B%7B%20unsubscribe_link%7Ce%20%7D%7D">Unsubscribe</a>` 

This happens even if I have base_url=None and disable_link_rewrites=True. My guess is it reads the link, checks that disable_link_rewrites is true and writes it back, but it's set to escape chars by default, thus escaping the mustaches.

Reason why it should be able to circumvent this is because it's way more performant to do the premailer transform once than with every e-mail when you're just using the template for simple things such as URLs and names

CraigRobertWhite:
Resolves #248 

#### Added
- `preserve_handlebar_syntax` option (Defaults to `False`)
   - Replaces any `"` between `="{{` and `}}"` in the `html` string (prior to being encoded) with `%22`.
      - This is to ensure that double-quotes within the handlebar syntax don't break when encoded.
      - This works because, as the below statement explains, everything with the `{{}}` (including `%22`) will be decoded after the rest of the string has been encoded.
   - Use `unquote()` to decode everything between every pair of `="%7B%7B` and `%7D%7D"` (the encoded versions of `{{` and `}}`).
   
### Motivations
- Like the author of #248, I too ran into an issue when attempting to transform a template with handlebar syntax in HTML attributes.
   - In my case, I was using the [Django templating language](https://docs.djangoproject.com/en/3.1/ref/templates/language/).

### Additional Context
- I just used the name `preserve_handlebar_syntax` because that's what @peterbe wrote in #248.

I really enjoy working with this library and would appreciate any feedback because I am using this library in other projects.

### DIFF
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -4,7 +4,7 @@ import os
 import re
 import warnings
 from collections import OrderedDict
-from urllib.parse import urljoin, urlparse
+from urllib.parse import urljoin, urlparse, unquote
 
 import cssutils
 import requests
@@ -116,6 +116,7 @@ class Premailer(object):
         disable_link_rewrites=False,
         preserve_internal_links=False,
         preserve_inline_attachments=True,
+        preserve_handlebar_syntax=False,
         exclude_pseudoclasses=True,
         keep_style_tags=False,
         include_star_selectors=False,
@@ -152,6 +153,7 @@ class Premailer(object):
         self.disable_link_rewrites = disable_link_rewrites
         self.preserve_internal_links = preserve_internal_links
         self.preserve_inline_attachments = preserve_inline_attachments
+        self.preserve_handlebar_syntax = preserve_handlebar_syntax
         self.exclude_pseudoclasses = exclude_pseudoclasses
         # whether to delete the <style> tag once it's been processed
         # this will always preserve the original css
@@ -518,6 +520,9 @@ class Premailer(object):
                 out = _cdata_regex.sub(
                     lambda m: "/*<![CDATA[*/%s/*]]>*/" % m.group(1), out
                 )
+            if self.preserve_handlebar_syntax:
+                while len(x := out.split('%7B%7B', 1)) > 1 and len(y := x[1].split('%7D%7D', 1)) > 1:
+                    out = x[0] + '{{' + unquote(y[0]) + '}}' + y[1]
             return out
 
     def _load_external_url(self, url):
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -308,6 +308,14 @@ class Premailer(object):
             else:
                 parser = etree.HTMLParser()
             stripped = html.strip()
+
+            if self.preserve_handlebar_syntax:
+                stripped = re.sub(
+                    r'="{{(.*?)}}"',
+                    lambda match: '="{{' + match.groups()[0].replace('"', '%22') + '}}"',
+                    stripped
+                )
+
             tree = etree.fromstring(stripped, parser).getroottree()
             page = tree.getroot()
             # lxml inserts a doctype if none exists, so only include it in
@@ -518,16 +526,11 @@ class Premailer(object):
                     lambda m: "/*<![CDATA[*/%s/*]]>*/" % m.group(1), out
                 )
             if self.preserve_handlebar_syntax:
-                while True:
-                    x = out.split("%7B%7B", 1)
-                    if len(x) > 1:
-                        y = x[1].split("%7D%7D", 1)
-                        if len(y) > 1:
-                            out = x[0] + "{{" + unquote(y[0]) + "}}" + y[1]
-                        else:
-                            break
-                    else:
-                        break
+                out = re.sub(
+                    r'="%7B%7B(.+?)%7D%7D"',
+                    lambda match: '="{{' + unquote(match.groups()[0]) + '}}"',
+                    out
+                )
             return out
 
     def _load_external_url(self, url):
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -518,8 +518,16 @@ class Premailer(object):
                     lambda m: "/*<![CDATA[*/%s/*]]>*/" % m.group(1), out
                 )
             if self.preserve_handlebar_syntax:
-                while len(x := out.split('%7B%7B', 1)) > 1 and len(y := x[1].split('%7D%7D', 1)) > 1:
-                    out = x[0] + '{{' + unquote(y[0]) + '}}' + y[1]
+                while True:
+                    x = out.split("%7B%7B", 1)
+                    if len(x) > 1:
+                        y = x[1].split("%7D%7D", 1)
+                        if len(y) > 1:
+                            out = x[0] + "{{" + unquote(y[0]) + "}}" + y[1]
+                        else:
+                            break
+                    else:
+                        break
             return out
 
     def _load_external_url(self, url):
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -4,6 +4,7 @@ import os
 import re
 import warnings
 from collections import OrderedDict
+from html import escape, unescape
 from urllib.parse import urljoin, urlparse, unquote
 
 import cssutils
@@ -312,9 +313,7 @@ class Premailer(object):
             if self.preserve_handlebar_syntax:
                 stripped = re.sub(
                     r'="{{(.*?)}}"',
-                    lambda match: '="{{'
-                    + match.groups()[0].replace('"', "%22")
-                    + '}}"',
+                    lambda match: '="{{' + escape(match.groups()[0]) + '}}"',
                     stripped,
                 )
 
@@ -530,7 +529,7 @@ class Premailer(object):
             if self.preserve_handlebar_syntax:
                 out = re.sub(
                     r'="%7B%7B(.+?)%7D%7D"',
-                    lambda match: '="{{' + unquote(match.groups()[0]) + '}}"',
+                    lambda match: '="{{' + unescape(unquote(match.groups()[0])) + '}}"',
                     out,
                 )
             return out
--- a/premailer/premailer.py
+++ b/premailer/premailer.py
@@ -312,8 +312,10 @@ class Premailer(object):
             if self.preserve_handlebar_syntax:
                 stripped = re.sub(
                     r'="{{(.*?)}}"',
-                    lambda match: '="{{' + match.groups()[0].replace('"', '%22') + '}}"',
-                    stripped
+                    lambda match: '="{{'
+                    + match.groups()[0].replace('"', "%22")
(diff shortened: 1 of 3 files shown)


peterbe:
The TravisCI testing has stopped working. We need to switch to GitHub Actions. 
I started https://github.com/peterbe/premailer/pull/253

CraigRobertWhite:
@peterbe https://github.com/peterbe/premailer/pull/252/commits/b470e16821954a2326c47bcf73f7e7ce354bbfc8

CraigRobertWhite:
I forgot that this package supported Python version pre-walrus. Going to fix asap. Sorry about that.

peterbe:
Yeah, I knew that was going to happen :)
The walrus syntax is neat and this new feature you're adding is opt-in but from a quick skim, if you can just rewrite in a "old school" way it would alleviate the problem for another couple of years.

CraigRobertWhite:
@peterbe I'm not crazy about this implementation without the walrus operator but it works. If you have any room for improvement, I'd definitely be willing to implement it. Also, I'm also totally cool with squash merging this PR, haha.

premailer/premailer.py
```diff
@@ -515,6 +517,17 @@ def transform(self, html=None, pretty_print=True, **kwargs):
...
+                while True:
+                    x = out.split("%7B%7B", 1)
+                    if len(x) > 1:
+                        y = x[1].split("%7D%7D", 1)
+                        if len(y) > 1:
+                            out = x[0] + "{{" + unquote(y[0]) + "}}" + y[1]
+                        else:
+                            break
+                    else:
+                        break
```
peterbe, line 530: Instead of the string-magic, how about purely looking for it in attributes that have quotation marks, as a regular expression?
E.g. 
```python
re.sub(
        r'="%7B%7B(.*?)%20%7D%7D"',
        lambda match: '="{{' + unquote(match.groups()[0]) + '}}"',
        html,
    )
```
Because `etree.tostring()` will always use `"` for attributes, right?

This way, you can be a little bit more certain it will only apply this hack on what appears to be HTML tag attributes.

premailer/tests/test_premailer.py
```diff
@@ -2984,3 +2984,57 @@ def test_pseudo_selectors_without_selector(self):
...
+    def test_preserve_handlebar_syntax_true(self):
+        """Demonstrate encoding of handlebar syntax with preservation enabled.
+
+        Original issue: https://github.com/peterbe/premailer/issues/248
+        """
+
+        html = """
+            <html>
+            <img src="{{ img_src }}">
+            <a href="{{ url }}"></a>
```
peterbe, line 3024: This second example doesn't exercise the importance of using `urllib.parse.unquote`. Perhaps...
```suggestion
            <a href="{{ url('some thing') }}"></a>
```

Note, I know next to nothing about Handlebar.

premailer/premailer.py
CraigRobertWhite: This is definitely a better way of doing it considering we know the contents of the HTML tag attributes are affected. I'm not too great at regex but your implementation definitely worked. Change included here: https://github.com/peterbe/premailer/pull/252/commits/03c0db4dc242a4b66f49facf55dfbb3a3a5bc575

premailer/tests/test_premailer.py
CraigRobertWhite: Including more complex content is a great idea. I discovered a problem when attempting to include `"` in the handlebars. My solution was to replace `"` within the handlebars with `%22` before being encoded so that when they were decoded they would come back out as normal `"`. Changes here: https://github.com/peterbe/premailer/pull/252/commits/03c0db4dc242a4b66f49facf55dfbb3a3a5bc575

CraigRobertWhite: I got the test data from https://docs.djangoproject.com/en/3.1/ref/templates/language/#string-literals-and-automatic-escaping because it seemed to be the most complex example of handlebar syntax that I could find.

CraigRobertWhite:
Sorry about the multiple commits for fixing formatting.

premailer/tests/test_premailer.py
```diff
@@ -2984,3 +2984,57 @@ def test_pseudo_selectors_without_selector(self):
...
+    def test_preserve_handlebar_syntax_false(self):
+        """Demonstrate encoding of handlebar syntax with preservation disabled.
+
+        Original issue: https://github.com/peterbe/premailer/issues/248
+        """
+
+        html = """
+            <html>
+            <img src="{{ data | default: 'Test.' }}">
+            <a href="{{ data | default: "Test." }}"></a>
```
peterbe, line 2997:
```suggestion
            <a href="{{ data | default: "Test & Testing" }}"></a>
```

peterbe: Or, 
```suggestion
            <a href="{{ data | default: "Test & <code>" }}"></a>
```

peterbe: Because, that's valid Handlebar code, right?

premailer/premailer.py
```diff
@@ -306,6 +308,16 @@ def transform(self, html=None, pretty_print=True, **kwargs):
...
             stripped = html.strip()
+
+            if self.preserve_handlebar_syntax:
+                stripped = re.sub(
+                    r'="{{(.*?)}}"',
+                    lambda match: '="{{'
+                    + match.groups()[0].replace('"', "%22")
+                    + '}}"',
+                    stripped,
+                )
```
peterbe, line 319: This needs to be explained. Because it's quite surprising. 

- One explanation that explains that Handlebar syntax might have its own characters that would normally get double encoded by `etree.tostring()`. Perhaps include an example from your unit test in the comment. (remember to not explain what the code does)
- One explanation for why this can't be done in the post-processing step on line 533.

Regarding the second one, I'm actually curious why it can't be done as part of that `unquote(...)` stuff you're doing around line 533.

premailer/tests/test_premailer.py
CraigRobertWhite: Good point. With this implementation, symbols like `&`, `<`, and `>` get encoded and `unquote()` doesn't successfully decode them back to their original symbols. I'm not quite sure where to begin trying to solve that. But, I'll be doing some research.

CraigRobertWhite: A solution that I've come up with is using `html.unescape()` to replace the named and character references with their original unicode characters.
In addition to discovering `unescape()`, I also found `escape()` which I've decided to replace my usage of `replace()` that replaced `"` with `%22` (which I'll attempt to justify in the below comment).

premailer/premailer.py
CraigRobertWhite: Regarding these explanations: You just want me to leave comments in the code to explain why the implementation was necessary and why a part of the implementation needed to occur prior to being encoded, right?

Point of clarity: The reason I thought I would have to escape the `"` characters prior to being encoded was that `etree.tostring()` looks at an attribute and treats its contents from the beginning `"` to the very next `"`. That's why:
```html
<a href="{{ data | default: "Test & <code>" }}"></a>
```
turns into
```html
<a href="%7B%7B%20data%20%7C%20default:%20" test>" }}"&gt;</a>
```
If we had to wait until after the encoding, it would seem that getting the original value back would be very difficult due to how `etree.tostring()` encodes attributes (as seen above).

peterbe: > Regarding these explanations: You just want me to leave comments in the code to explain why the implementation was necessary and why a part of the implementation needed to occur prior to being encoded, right?

Exactly. 

I usually think, "Does this code make me have to stop and think? If not today, what about when I (or someone else!) read this code in 6 months."

premailer/tests/test_premailer.py
```diff
@@ -2984,3 +2984,57 @@ def test_pseudo_selectors_without_selector(self):
...
+    <img src="%7B%7B%20data%20%7C%20default:%20'Test%20&amp;%20&lt;code&gt;'%20%7D%7D">
+    <a href="%7B%7B%20data%20%7C%20default:%20" test>" }}"&gt;</a>
+    </body>
+</html>
+        """
+        p = Premailer(html)
+        result_html = p.transform()
+        compare_html(expect_html, result_html)
+
+    def test_preserve_handlebar_syntax_true(self):
```
peterbe, line 3015: Can you please collapse these two tests into one. The first one is kinda crazy in that it doesn't actually test anything. It just does what it would do if you don't apply the functionality :)
But I quite like the effect that it demonstrates, so I think it's worth keeping. 

If you just had _1_ test for both `preserve_handlebar_syntax=True` and `preserve_handlebar_syntax=False` you wouldn't need to repeat the input string. And you wouldn't need to repeat the comment about which original issue this revolves around.

peterbe:
Hey, as a general comment the tests look great (apart from the note about making the 2 unit tests into 1). 
I don't know if I understand the `unquote` and `unescape` and all that but the tests clearly prove that what comes out of the other end is sane and desired. I expected it to be a bit weird because HTML + Handlebar isn't HTML so hacking around it is the only sensible option. I've seen it many times with Django Templates and Jinja templates. For example, some IDEs try to syntax highlight it as HTML when it's actually something entirely different. 

If you can just heed the refactoring nit about the unit tests and write some _why-comments_ about the stuff before and after `.tostring()` then I'll approve and make a release.

premailer/tests/test_premailer.py
CraigRobertWhite: Sure. Here is the commit where I combined the two tests: https://github.com/peterbe/premailer/pull/252/commits/0dc619286a594bab7a1f716d74c094542561718a

As a side-note, in defense of having two tests:
- The first test, with the setting turned off, tests a case in which using handlebar syntax breaks the encoding. While I'm not sure it warranted its own test, I felt that it was good to show to the user (hence including in the 1, combined test). And, because I thought it warranted being included in a test, I followed the ["Arrange Act Assert"](http://wiki.c2.com/?ArrangeActAssert) test pattern.
   - Combining the two tests breaks that testing pattern because it requires the test to "Act" twice.

All that to say, I'm totally cool with combining them, I just wanted to say why I had two to begin with :)

premailer/premailer.py
CraigRobertWhite: Working on this now

CraigRobertWhite: I might have been a little too verbose with my comments but I'd rather do that than have people not understand what's going on: https://github.com/peterbe/premailer/pull/252/commits/8426f81e1e8200ac7ac80584036ea3f8c46fa799 Let me know if you want me to change anything.

peterbe (approved):

PR #255, opened 2021-06-01
prep for v 3.9.0

peterbe:


johan-bjareholt:
I have a Makefile using premailer and used the following ugly hack to support both 3.9.0 and below

```
# the --allow-loading-external-files argument was added in premailer version
# 3.9.0 (released 2021-06-01) and is required for html to build with the new
# version
PREMAILER_ARGS = $(shell $(PYTHON) -c "import premailer; version=premailer.__version__.split('.'); args='--allow-loading-external-files' if int(version[0])>=3 and int(version[1])>=9 else ''; print(args)")
```

PR #265, opened 2021-10-13
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268 .


browniebroke:
I was surprised to se that it actually [still works in Python 3.11](https://github.com/browniebroke/premailer/actions/runs/3740775842/jobs/6349545967#step:6:78)... It turns out this removal was reverted and deferred until Python 3.12: https://github.com/python/cpython/commit/b50322d20337ca468f2070eedb051a16ee1eba94

Would be nice to get it fixed anyway 👍🏻

PR #288, opened 2022-12-20
Add official support for Python 3.10 and 3.11

browniebroke:
The main blocker were the test runner, nose which is no longer maintained and has several open issues for recent versions of Python:

- https://github.com/nose-devs/nose/issues/1099
- https://github.com/nose-devs/nose/issues/1118
- https://github.com/nose-devs/nose/issues/1116
- https://github.com/nose-devs/nose/issues/1122

I've replaced nose test runner by pytest (which is widely used in the Python community), let me know if you disagree with this choice, and if you prefer another alternative.

Related: #265


.github/workflows/python.yml
```diff
@@ -7,10 +7,11 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
browniebroke, line 10: It's no longer possible to install Python 3.5 and 3.6 on ubuntu-latest which now points to 22.04. They are both EOL so it's probably time to drop support for them, but that feels out of scope of this change.

premailer/tests/test_premailer.py
```diff
@@ -23,6 +23,32 @@
...
+    Copied from Nose.
+    """
+    if not a == b:
+        raise AssertionError(msg or "%r != %r" % (a, b))
+
+
+def assert_raises(exc_class, func, *args):
+    """Compatibility with Nose API for pytest."""
+    with pytest.raises(exc_class):
+        func(*args)
```
browniebroke, line 49: These provide compatibility with Nose APIs. I can replace their usage in this file with direct pytest equivalent if you prefer.

AndreyObviously:
@peterbe Is this planned for release?

premailer/tests/test_premailer.py
peterbe: > These provide compatibility with Nose APIs. I can replace their usage in this file with direct pytest equivalent if you prefer.

A search-and-replace is probably better.

peterbe:
> @peterbe Is this planned for release?

No. Not until all the tests are updated to support modern versions.

browniebroke:
I no longer use this project so have little incentive to progress this further.

> No. Not until all the tests are updated to support modern versions.

What do you mean by that? Can you give a version range to target?

browniebroke:
Tried to get this more up to date but it looks like it needs a bit more work as one test is now failing (I presume because a transitive dependency has changed its behaviour)...

I don't plan on updating this further but to whoever is motivated, feel free to reuse this 😊

peterbe:
> I no longer use this project so have little incentive to progress this further.
> 

I don't use it either.

> > No. Not until all the tests are updated to support modern versions.
> 
> What do you mean by that? Can you give a version range to target?

>=3.12 perhaps?

@AndreyObviously If you need modern versions of this, perhaps take a look at the work by @browniebroke and see if you can create a new fresh PR.

PR #292, opened 2023-08-30
Include capitalize_float_margin in Premailer initialization

Eboreg:
The "--capitalize-float-margin" command line argument was not forwarded to Premailer, now it is.


NavinPa (approved):
( . )

